PR #4, opened 2020-01-16
Tag has now typed domNode

jwstegemann:
working, but I do not know, if it is really needed.
The type-parameters will add quite some bytes to the size of the resulting java-script...


jwstegemann:
Merged with functional state management

PR #6, opened 2020-01-23
Functional seq

jwstegemann:


jwstegemann:
not realy tested yet but I like it. Seq eliminated, too. Have a look at the demo... ;-)

src/main/kotlin/demo.kt
```diff
@@ -34,7 +42,19 @@ fun main() {
                     ActionData(it.clientX, it.clientY)
                 }
             }
-
+            ul {
+                seq.each().map { s: String ->
```
jamowei: Can you add an example with deletion of an specific element in the list. For example when clicking it.
In my opinion there is missing an index number or it is possible to use an object reference?

src/main/kotlin/io/fritz2/binding/seq.kt
```diff
@@ -1,54 +1,46 @@
...
 
+private suspend inline fun <T> accumulate(accumulator: Pair<List<T>, List<T>>, newValue: List<T>): Pair<List<T>, List<T>> = Pair(accumulator.second, newValue)
+
+@ExperimentalCoroutinesApi
+@FlowPreview
+fun <T> Store<List<T>>.each(): Flow<Patch<T>>  =
+    data.scan(Pair(emptyList<T>(), emptyList<T>()), ::accumulate).flatMapConcat(::compare)
+
 
 //TODO: flatmap needed?
```
jamowei: Yes, I think we need it in case of matrices (like lists of lists). But it is a special case...

src/main/kotlin/io/fritz2/binding/seq.kt
```diff
@@ -1,54 +1,46 @@
...
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.channels.ConflatedBroadcastChannel
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.asFlow
-import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.*
 import kotlinx.coroutines.flow.map
 
 data class Patch<out T>(val from: Int, val that: List<T>, val replaced: Int)
```
jamowei: Maybe a bit more comments here!?

src/main/kotlin/io/fritz2/binding/seq.kt
```diff
@@ -1,54 +1,46 @@
...
-        last = n
+    //FIXME: better performance without range?
+    for (i in 0 until size2Compare) {
+        //TODO: batch changed items in a row to one patch
+        //console.log("### Seq: inner comparing: $i -> ${last[i]} to ${n[i]}")
+        if (oldValue[i] != newValue[i]) channel.send(Patch(i, listOf(newValue[i]), 1))
     }
 }
 
+private suspend inline fun <T> accumulate(accumulator: Pair<List<T>, List<T>>, newValue: List<T>): Pair<List<T>, List<T>> = Pair(accumulator.second, newValue)
```
jamowei: Maybe add a comment here!?

src/main/kotlin/demo.kt
```diff
@@ -24,6 +27,9 @@ fun main() {
         val addItem = Handler<Any> { list, _ ->
             list + "yet another item"
         }
+        val deleteItem = Handler<Int> { list, i ->
```
jamowei: this would my first try to delete specific elements from list...?

src/main/kotlin/demo.kt
```diff
@@ -43,10 +49,11 @@ fun main() {
                 }
             }
             ul {
-                seq.each().map { s: String ->
+                seq.each().mapIndexed{ i:Int, s: String ->
```
jamowei: new method or the map-method get a second parameter with index!?

src/main/kotlin/demo.kt
```diff
@@ -43,10 +49,11 @@ fun main() {
...
                     html {
-                        li {
+                        button {
                             +s
+                            seq.deleteItem <= clicks.map { console.log(i); i }
```
jamowei: in console it looks good...?

src/main/kotlin/io/fritz2/binding/seq.kt
```diff
@@ -43,4 +43,10 @@ fun <T> Store<List<T>>.each(): Flow<Patch<T>>  =
 fun <T, X> Flow<Patch<T>>.map(mapper: (T) -> X): Flow<Patch<X>> =
     this.map {
         Patch(it.from, it.that.map(mapper), it.replaced)
-    }
+}
+
+//FIXME: not working properly...?
+fun <T, X> Flow<Patch<T>>.mapIndexed(mapper: (Int, T) -> X): Flow<Patch<X>> =
```
jamowei: not sure if its correct!?

src/main/kotlin/demo.kt
jamowei: Ok, I added my own simple example, but its not working right now... :/

jwstegemann: I think using an index is not a good idea. just use the reference. In the future we have to use identification via id / idProvider-interface or something like this.

src/main/kotlin/io/fritz2/binding/seq.kt
jwstegemann: ok, but why would we flatten the matrix. A row surely will have to be rendered differently than a column, for example.

src/main/kotlin/io/fritz2/binding/seq.kt
jwstegemann: I just saw that a branched the new structure from this branch not the master (my fault), so there is no merging needed... ;-)

src/main/kotlin/io/fritz2/binding/seq.kt
jwstegemann: why? ok, have commented, but in an actual branch...

src/main/kotlin/demo.kt
jwstegemann: definitely a second method. different semantics. but I still do not think, it is a good idea.

src/main/kotlin/io/fritz2/binding/seq.kt
jwstegemann: as I said above...

src/main/kotlin/demo.kt
jamowei: Ok, I fixed the example. My mistake. Now its working perfectly :)

src/main/kotlin/io/fritz2/binding/seq.kt
```diff
@@ -43,10 +43,4 @@ fun <T> Store<List<T>>.each(): Flow<Patch<T>>  =
 fun <T, X> Flow<Patch<T>>.map(mapper: (T) -> X): Flow<Patch<X>> =
     this.map {
         Patch(it.from, it.that.map(mapper), it.replaced)
-}
-
-//FIXME: not working properly...?
-fun <T, X> Flow<Patch<T>>.mapIndexed(mapper: (Int, T) -> X): Flow<Patch<X>> =
-    this.map {
-        Patch(it.from, it.that.mapIndexed(mapper), it.replaced)
 }
```
jamowei: removed mapIndexed, cause not needed

PR #7, opened 2020-01-24
Add w3c specs

jamowei:
First draft finished. Please take a look ;-)


jamowei:
Ok, I hope I have added the most of the html5 elements...
Next step is to write some tests. Has anyone a good idea for it?

jwstegemann:
great. jan wanted to write a todomvc-example. he will have to use quite a lot of elements...

core/src/main/kotlin/io/fritz2/dom/html/elements.kt
```diff
@@ -7,24 +7,71 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
...
+class Br : Tag<HTMLBRElement>("br")
 
-//FIXME: use correct type for domNode - HtmlButtonElement here
 @ExperimentalCoroutinesApi
 @FlowPreview
+/**
+ * Exposes the JavaScript [HTMLButtonElement](https://developer.mozilla.org/en/docs/Web/API/HTMLButtonElement) to Kotlin
+ */
 class Button : Tag<HTMLButtonElement>("button"), WithText<HTMLButtonElement> {
     var autofocusFlow: Flow<Boolean> by AttributeDelegate
```
jwstegemann: why is this calls autofocusFlow? same for next line?

jamowei: Oh, I didn't saw that. Thanks for reporting. Fix that in next commit.

jamowei:
Renamed events. Therefore I must use 'on'<event-name> otherwise I get conflicts with some html attributes, e.g. "search". I also added a new syntax for handling the events. Please take a look. It is more the js-way and you can read the code from left to right. I personally liked it.

before:
```kotlin
store.addADot <= onclick.map {
    ActionData(it.clientX, it.clientY)
}
```
now:
```kotlin
onclick.map {
    ActionData(it.clientX, it.clientY)
} >= store.addADot
```

Actual there are both options possible. What do you think?

jamowei:
I don't add extra tests for all html elements. Some of the elements getting tested by the todomvc example.

jamowei:
Do we need some more convenience functions/events?
Like "change" event with getting the string value directly from <input>, <select> and <textarea>

jwstegemann:
I think >= is totally confusing as an operator. In this case I would prefer ..
We have to decide for one way. It would be too confusing, if there are two different ways to achieve something as simple as event-handling..
Handler is the wrong place to convert the event-type. The handler belongs to the store and should not be limited to be used with a specific event source. The same action can be requested by clicking on a button or pressing return for example...
I like the handler <= event syntax because you get the two most important information at a glance, even with complex mappings.

jwstegemann:
I would prefer to go with the plural-form than. onEvent is too close to JavaScript since there is a fundamental difference by using streams of events instead of just defining a handler. I do not think, that it is a good idea to hide this from the user..

jamowei:
> I would prefer to go with the plural-form than. onEvent is too close to JavaScript since there is a fundamental difference by using streams of events instead of just defining a handler. I do not think, that it is a good idea to hide this from the user..

Okay, I will turn it back. Then we have "focuss" for example.

jamowei:
> I think >= is totally confusing as an operator. In this case I would prefer ..
> We have to decide for one way. It would be too confusing, if there are two different ways to achieve something as simple as event-handling..
> Handler is the wrong place to convert the event-type. The handler belongs to the store and should not be limited to be used with a specific event source. The same action can be requested by clicking on a button or pressing return for example...
> I like the handler <= event syntax because you get the two most important information at a glance, even with complex mappings.

Ok, was an idea. I will turn it also back.

PR #9, opened 2020-01-26
stores for deep structures

jwstegemann:
what do you think?


examples/nestedmodel/src/main/kotlin/io/fritz2/examples/nestedmodel/nestedmodel.kt
```diff
@@ -0,0 +1,75 @@
...
+
+
+
+@ExperimentalCoroutinesApi
+@FlowPreview
+fun main() {
+
+    val model = Store(Outer(Inner("hello"), "world"))
+    val outerValue = model.sub(Outer.valueLens)
+    val innerValue = model.sub(Outer.innerLens).sub(Inner.valueLens)
```
jamowei: why not use: `val innerValue = model.sub(Outer.innerLens + Inner.valueLens)`?
Is it not easier to generate?

examples/nestedmodel/src/main/kotlin/io/fritz2/examples/nestedmodel/nestedmodel.kt
```diff
@@ -0,0 +1,75 @@
...
+}
+
+data class Inner(val value: String) {
+    companion object {
+        val valueLens = object : Lens<Inner, String> {
+            override fun get(parent: Inner): String = parent.value
+            override fun set(parent: Inner, value: String): Inner = parent.copy(value = value)
+        }
+    }
+}
```
jamowei: Does it work with seqs when other pull-request is completed?

core/src/main/kotlin/io/fritz2/binding/substores.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun get(parent: P): T
+    fun set(parent: P, value: T): P
+    fun map(parent: P, mapper: (T) -> T): P = set(parent, mapper(get(parent)))
+
+
+    operator fun <X> plus(other: Lens<T,X>): Lens<P,X> = object : Lens<P,X> {
+        override fun get(parent: P): X = other.get(this@Lens.get(parent))
+        override fun set(parent: P, value: X): P = this@Lens.set(parent, other.set(this@Lens.get(parent), value))
+    }
+}
```
jamowei: Maybe add some generic id generation here?

examples/nestedmodel/src/main/kotlin/io/fritz2/examples/nestedmodel/nestedmodel.kt
jamowei: I tried it and it works as expected :+1:

core/src/main/kotlin/io/fritz2/binding/substores.kt
```diff
@@ -0,0 +1,39 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+
+interface Lens<P,T> {
+    fun get(parent: P): T
+    fun set(parent: P, value: T): P
+    fun map(parent: P, mapper: (T) -> T): P = set(parent, mapper(get(parent)))
```
jamowei: When do I use the map function?

examples/nestedmodel/src/main/kotlin/io/fritz2/examples/nestedmodel/nestedmodel.kt
jwstegemann: It's perfectly valid. I just wanted to test, if two successive calls to sub work...

examples/nestedmodel/src/main/kotlin/io/fritz2/examples/nestedmodel/nestedmodel.kt
jwstegemann: not yet. working on it.

core/src/main/kotlin/io/fritz2/binding/substores.kt
jwstegemann: definitely! I think this should be done in code-generation by just adding an ´´´val id = "${parent.id}.<generated id>"```

jamowei:
Hi, first of all its looking good. For me its a bit confusing that you must use diffrent methods for similar things:

```kotlin
seq.eachStore().mapItems { ... }.bind()
seq.each().mapItems { ... }.bind()
```
Also it is for me confusing that I cant use
```kotlin
seq.eachStore().map { ... }.bind()
seq.each().map { ... }.bind()
```
cause of `map(...)` coming from Flow like all the other transforming methods, which I can't use in this place. Therefore my autocomplete is not much helpful and confusing.

I would appreciate then to use it like this:
```kotlin
seq.each().transform().html { ... }.bind()
seq.each().transform().html { ... }.bind()
```
where it dosen't matter if it comes from `AbstractStore` or `Store` and `transform()` stands for all tranformation functions like map, flatMap, filter, ...
The `html {...}` block then containing all the rendered html elements with the data from the seqs.

What if I want to merge two seqs from different `Stores` or different places in the same `Store` together with `s1.flatMap { e1 -> s2.map { e2 -> .... }}` construct? How would I do it?

jamowei:
Why is the interface `withId` written with small letter?
```kotlin
interface withId {
    val id: String
}
```
Maybe there is a better way of using it like this?
```kotlin
data class Entry(val num: Int, val text: String) : withId {
    override val id: String
        get() = num.toString()
}
```
My idea is to use an abstract class like this:
```kotlin
abstract class WithId(val id: String)
data class Entry(val num: Int, val text: String): WithId(num.toString())
```

jwstegemann:
> ```kotlin
> abstract class WithId(val id: String)
> data class Entry(val num: Int, val text: String): WithId(num.toString())
> ```
Good idea. Feel free to implement... ;-)

jwstegemann:
> What if I want to merge two seqs from different `Stores` or different places in the same `Store` together with `s1.flatMap { e1 -> s2.map { e2 -> .... }}` construct? How would I do it?

Since the Store is the single point of truth you have to change the value there.
You can combine Lists the way you wrote, but you cannot combine Seqs, because they are streams of Patches and it makes no sense to mix patches of one Seq with the patches of another one, especially not, if you want to do two-way-binding downstream.

jwstegemann:
> cause of `map(...)` coming from Flow like all the other transforming methods, which I can't use in this place. Therefore my autocomplete is not much helpful and confusing.
>
That is the problem. Since ´map` is defined in `Flow` it has a higher priority than any extension we could provide. Therefore we need to use a different name. Maybe it is not that bad, since you can see from it, that you do not just map the value, but only a part of the value (only the items within the `Patch`)  

`each()`and `eachStore` have completely different semantics. While `each()` gives you a `Seq`of the Item-Type directly (which you had to unwrap in bsui5 to use it), `eachStore()` gives you a `Seq` of Stores of the Item-Type. Therefore you have to use the later one, when you need two-way-databinding or sub-flows (to not render the whole item again, when only one attribute changes for example).

 
> ```kotlin
> seq.each().transform().html { ... }.bind()
> seq.each().transform().html { ... }.bind()
> ```
> 
> where it doesn't matter if it comes from `AbstractStore` or `Store` and `transform()` stands for all transformation functions like map, flatMap, filter, ...
> The `html {...}` block then containing all the rendered html elements with the data from the seq entries.
You can use every function from Flow an the `Seq`, since it is just a `Flow<Patch<...>>`. Your just have to map the whole Patch, of filter the Patch, or whatever. mapItems is just a convenience-method. We should offer more for ´filter`, etc.
You would need something like a MonadTransformer to implement it just one for alle the functions, but I have no idea how to do that in Kotlin, yet.

jwstegemann:
> ```kotlin
> ```kotlin
> abstract class WithId(val id: String)
> data class Entry(val num: Int, val text: String): WithId(num.toString())
> ```
> 
> 
> 
> Good idea. Feel free to implement... ;-)
> ```

Putting the if as Val in an abstract class isn’t a good idea, since it is not used by the data class features than anymore (no getter/setter, equals, toStribg, copy!). But you can write your first suggestions even easier by
`Data class hugo(Val num:Int. .. override Val Id: String = num.String())`

PR #10, opened 2020-02-01
Added id & class attribute + methods for data-attributes

jamowei:


jamowei:
Added missing attributes

core/src/main/kotlin/io/fritz2/dom/tag.kt
```diff
@@ -29,4 +30,11 @@ abstract class Tag<out T : Element>(tagName: String, override val domNode: T = w
     fun <X : Element> Flow<Tag<X>>.bind(): SingleMountPoint<WithDomNode<Element>> = DomMountPoint(this, domNode)
 
     fun <X : Element> Flow<Patch<Tag<X>>>.bind(): MultiMountPoint<WithDomNode<Element>> = DomMultiMountPoint(this, domNode)
+
+    operator fun <T> T.not() = Const(this)
```
jamowei: I put it in here cause then no extra import is needed. So it's a bit easier to use ;)

jwstegemann:
And now: an AttributeMountPoint for Seq (classes)

Jan <notifications@github.com> schrieb am Mo. 3. Feb. 2020 um 22:39:

jamowei:
> And now: an AttributeMountPoint for Seq (classes) Jan <notifications@github.com> schrieb am 
Done

PR #11, opened 2020-02-03
Validation

jamowei:
Have a problem with the validation interface, because it can't hold a value like Flow<Message>...


jamowei:
What do you think?

PR #13, opened 2020-02-06
Optimize validation

jamowei:
Added and fixed some stuff in validation.

### DIFF
--- a/core/src/main/kotlin/io/fritz2/binding/validation.kt
+++ b/core/src/main/kotlin/io/fritz2/binding/validation.kt
@@ -9,42 +9,25 @@ import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
 
-interface WithSeverity: withId {
-    val severity: Severity
-}
-
-enum class Severity {
-    Info,
-    Warning,
-    Error,
-    Fatal,
+interface Failable: withId {
+    fun isFail(): Boolean
 }
 
 @FlowPreview
 @ExperimentalCoroutinesApi
-abstract class Validator<D, M: WithSeverity, T> {
+abstract class Validator<D, M: Failable, T> {
 
     internal val channel = ConflatedBroadcastChannel<List<M>>()
     val msgs = channel.asFlow().distinctUntilChanged()
 
     abstract fun validate(data: D, metadata: T): List<M>
 
-    open fun validPredicate(msg: M): Boolean {
-        return msg.severity < Severity.Error
-    }
-
-    fun isValid(): Flow<Boolean> {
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/io/fritz2/binding/validation.kt
```diff
@@ -46,7 +52,7 @@ interface Validation<D, M: WithSeverity, T> {
         val messages = validator.validate(data, metadata)
         println(messages)
         validator.channel.offer(messages)
-        return messages.all { m -> validator.isValid(m) }
+        return messages.none { m -> !validator.validPredicate(m) }
```
jwstegemann: I think a better way would by using `messages.any(!validator::invalidPredicate)`

core/src/main/kotlin/io/fritz2/binding/validation.kt
```diff
@@ -26,13 +29,16 @@ abstract class Validator<D, M: WithSeverity, T> {
...
         return msg.severity < Severity.Error
     }
 
     fun isValid(): Flow<Boolean> {
-        //TODO optimize
-        return msgs.map { list -> list.none {m -> !isValid(m)}}
+        return msgs.map { list -> list.none {m -> !validPredicate(m)}}
+    }
+
+    fun isNotValid(): Flow<Boolean> {
```
jwstegemann: a lazy property would be better imho, so there will be at most one instance of the mapped flow

core/src/main/kotlin/io/fritz2/binding/validation.kt
```diff
@@ -26,13 +29,16 @@ abstract class Validator<D, M: WithSeverity, T> {
 
     abstract fun validate(data: D, metadata: T): List<M>
 
-    fun isValid(msg: M): Boolean {
+    open fun validPredicate(msg: M): Boolean {
```
jwstegemann: since this is the only information we need about a message, wouldn't it be a good idea to put just a function `isInvalid` in the interface and no property at all?

examples/validation/src/main/kotlin/io/fritz2/examples/validation/validation.kt
```diff
@@ -37,6 +38,7 @@ fun main() {
     val store = object : RootStore<String>(""), Validation<String, ValMsg, String> {
         override val validator = EMailValidator
 
+
         val updateWithValidation = Handler<String> { data, newData ->
```
jwstegemann: Should we put this (as lazy) inside the Validation-interface, so every validated store inherits if and it can be used without any code?

jamowei:
> Looks good, but has conflicts... (cleanup in nestedmodel.kt ;-))

Oh my bad, now its fixed

PR #14, opened 2020-02-12
Remote

jwstegemann:
remote access


examples/remote/src/main/kotlin/io/fritz2/examples/remote/remote.kt
```diff
@@ -20,9 +20,9 @@ fun main() {
 
     val store = object : RootStore<String>("start") {
 
-        val callApi = handle<String> { _, _ ->
-            URL("https://api.github.com").get() andThen update
-        }
+        val callApi = apply { s : String ->
```
jwstegemann: what do you think?

core/src/main/kotlin/io/fritz2/remote/remote.kt
```diff
@@ -2,32 +2,196 @@ package io.fritz2.remote
...
+        referrer = referrer,
+        referrerPolicy = referrerPolicy,
+        mode = mode,
+        credentials = credentials,
+        cache = cache,
+        redirect = redirect,
+        integrity = integrity,
+        keepalive = keepalive,
+        window = window
+    )
```
jamowei: Why these functions are public? They are a bit confusing when using the API and I think only need internal!?

PR #24, opened 2020-03-03
throw exception when tried to render multiple root tags

### LINKED ISSUE #21: html renders only one child-element
calling `html` to build a tree of elements, you can add more than one child to the root-`html` in perfectly valid code. 
Since it is later bound using a `SingleMountPoint` only the last added child is visible in the DOM in the browser, because the others are simply overridden. This is not intuitive.

jamowei:
fixes #21

### DIFF
--- a/core/src/main/kotlin/io/fritz2/dom/html/html.kt
+++ b/core/src/main/kotlin/io/fritz2/dom/html/html.kt
@@ -9,8 +9,20 @@ import org.w3c.dom.Element
 @FlowPreview
 fun <E : Element> html(content: HtmlElements.() -> Tag<E>) =
     content(object : HtmlElements {
+
+        var alreadyRegistered: Boolean = false
+
         override fun <X : Element, T : Tag<X>> register(element: T, content: (T) -> Unit): T {
-            content(element)
-            return element
+            if (!alreadyRegistered) {
+                content(element)
+                alreadyRegistered = true
+                return element
+            } else {
+                throw MultipleRootElementsException(
+                    "Don't use multiple root elements/tags in the html{...} context of your component! That's why rendering is not working."
+                )
+            }
         }
     })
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/io/fritz2/dom/html/html.kt
```diff
@@ -9,8 +9,20 @@ import org.w3c.dom.Element
 @FlowPreview
 fun <E : Element> html(content: HtmlElements.() -> Tag<E>) =
     content(object : HtmlElements {
+
+        var alreadyRegistered: Boolean = false
+
         override fun <X : Element, T : Tag<X>> register(element: T, content: (T) -> Unit): T {
-            content(element)
-            return element
+            if (!alreadyRegistered) {
```
jwstegemann: Why not just
```if (alreadyRegistered) throw ... ```

File: core/src/main/kotlin/io/fritz2/dom/html/html.kt [hunk 2]
```diff
@@ -9,8 +9,20 @@ import org.w3c.dom.Element
...
+                content(element)
+                alreadyRegistered = true
+                return element
+            } else {
+                throw MultipleRootElementsException(
+                    "Don't use multiple root elements/tags in the html{...} context of your component! That's why rendering is not working."
```
jwstegemann: Suggestion:
You can have only one root-tag per html-context.

jamowei: Ok, I have changed it.

core/src/main/kotlin/io/fritz2/dom/html/html.kt
jamowei: You are right! Its better to read.

PR #30, opened 2020-03-16
Event listener

jamowei:
first version


core/src/main/kotlin/io/fritz2/dom/listener.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.w3c.dom.HTMLSelectElement
+import org.w3c.dom.HTMLTextAreaElement
+import org.w3c.dom.events.Event
+import org.w3c.dom.events.KeyboardEvent
+import org.w3c.files.FileList
+
+/**
+ * [Action] contains the fired [Event] and targeting [Element]
+ */
+data class Action<E: Event, X: Element>(val event: E, val target: X)
```
jwstegemann: I think the name `Action` is wrong here. An action is, what you map your Event/Target/value to later on.
Also I do not think, that we need both seperated target and event. We can access target whenever needed from event and cast it. All we need is the type-information about the target.

core/src/main/kotlin/io/fritz2/dom/listener.kt
```diff
@@ -0,0 +1,81 @@
...
+ * the [Event] with [events] as Flow or
+ * the targeting [Element] with [targets] also as Flow back.
+ * If you don't need either the [Event] or [Element] you can call the [Listener]
+ * directly (e.g. `clicks()`) to get an Flow of [Unit] instead.
+ */
+@FlowPreview
+@ExperimentalCoroutinesApi
+inline class Listener<E: Event, X: Element>(val actions: Flow<Action<E, X>>) {
+    operator fun invoke(): Flow<Unit> = actions.map { Unit }
+    fun events(): Flow<E> = actions.map { it.event }
```
jwstegemann: I would stay with just one "basic" flow: events
This is the callbackFlow of the EventHandler. Without any mapping.

core/src/main/kotlin/io/fritz2/dom/listener.kt
```diff
@@ -0,0 +1,81 @@
...
+    fun targets(): Flow<X> = actions.map { it.target }
+}
+
+/**
+ * Gives you the new value as [String] from the targeting [Element]
+ */
+@FlowPreview
+@ExperimentalCoroutinesApi
+fun Listener<Event, HTMLInputElement>.value(): Flow<String> = targets().map { it.value }
```
jwstegemann: should be named `values`because it is a Flow.
Same for all the others

core/src/test/kotlin/io/fritz2/binding/events.kt
```diff
@@ -25,7 +25,7 @@ class EventTests {
         val store = object : RootStore<String>("start") {
             var countHandlerCalls = 0
 
-            val addADot = handle <Any> { model, _ ->
```
jwstegemann: Great! Looks much less confusing for newbies...

core/src/test/kotlin/io/fritz2/binding/seq.kt
```diff
@@ -80,39 +80,39 @@ class SeqTests {
 
                 button("replaceList") {
                     +"replaceList"
-                    store.replaceList <= clicks
+                    store.replaceList <= clicks()
                 }
                 button("addAtBeginning") {
                     +"addAtBeginning"
-                    store.addAtBeginning <= clicks
+                    store.addAtBeginning <= clicks()
```
jwstegemann: Not sure if invoke is a good idea. It looks good when you use it, but will it be clear that it is `clicks()` but `clicks.map {...}`?

core/src/main/kotlin/io/fritz2/dom/html/events.kt
```diff
@@ -1,21 +1,22 @@
...
-@ExperimentalCoroutinesApi
-@FlowPreview
-open class EventType<T>(val name: String) {
-    open fun extract(event: Event): T = event.unsafeCast<T>()
+/**
+ * [EventType] contains the javascript related [name] of an [Event]
+ * and has a method [extract] to cast an [Event] to its actual event-type.
+ */
+class EventType<T: Event>(val name: String) {
+    fun extract(event: Event): T = event.unsafeCast<T>()
```
jwstegemann: function call is unnecessary now. Just include it where it is called now `offer(....)`.

PR #31, opened 2020-03-17
Refactor binding methods

jamowei:
Here is the new idea for changing the binding methods.


jwstegemann:
is this resolved by [tidy up syntax for text, attr, const](https://github.com/jwstegemann/fritz2/commit/444ccf31e9c919e49ce676787b41488fd73b9cab)?

PR #39, opened 2020-03-25
TodoMVC

jwstegemann:
... "complete "example


jwstegemann:
Tests have to be adjusted to baseClass es default param (instead of id)

PR #40, opened 2020-03-30
Todomvc

jwstegemann:
error corrections


jwstegemann:
now without error-messages

PR #41, opened 2020-03-31
Improve examples

jamowei:


jamowei:
I have updated the gettingstarted example and added a new github build file for the deployment to github-pages

PR #52, opened 2020-04-06
Better diff

jwstegemann:
using Myer's diff algorithm for lists


jwstegemann:
seems to work (tests and examples)
please have a look...

jwstegemann:
No annotations necessary in core (have a look at build.gradle.kt!)

PR #58, opened 2020-04-14
Improve examples

jamowei:
Reworked all examples. Please check if something not right.


.gitignore
```diff
@@ -5,4 +5,4 @@
 **/.idea/
 **/gradle/
 **/src/generated/
-/examples/nestedmodel/model/src/main/kotlin/io/fritz2/examples/nestedmodel/GeneratedOptics.kt
```
jwstegemann, line 8: This should be generated in model-subproject...

core/src/main/kotlin/io/fritz2/binding/store.kt
```diff
@@ -76,3 +113,28 @@ open class RootStore<T>(initialData: T, override val id: String = "", bufferSize
 
     override fun <X> sub(lens: Lens<T, X>) = SubStore(this, lens, this, lens)
 }
+
+
+interface LensId<T> {
```
jwstegemann, line 118: I am not sure, if the name explains, what it does. Maybe "ModelId"?

core/src/main/kotlin/io/fritz2/dom/mount.kt
```diff
@@ -80,10 +81,19 @@ class AttributeMountPoint(val name: String, upstream: Flow<String>, val target:
...
+        when (value) {
+            is String -> {
+                target?.unsafeCast<HTMLInputElement>()?.value = value
+            }
+            is Double -> {
+                target?.unsafeCast<HTMLInputElement>()?.valueAsNumber = value
+            }
+            is Date -> {
+                target?.unsafeCast<HTMLInputElement>()?.valueAsDate = value
+            }
```
jwstegemann, line 95: missing else?

jwstegemann:
really great work!
it is good looking and demonstrates perfectly, how it works and feels...

jwstegemann:
Maybe add an examples badge on top of readme?

core/src/main/kotlin/io/fritz2/binding/store.kt
jamowei: Renamed in ModelId

core/src/main/kotlin/io/fritz2/dom/mount.kt
jamowei: new concept

PR #76, opened 2020-06-01
Readme -> English improvements

EvaBS:
Bitte kurz prüfen. Habe u.a. von "Ich" auf "Wir" geändert (bei Inspiration).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@
 [![Docs](https://img.shields.io/badge/docs-wiki-blue)](https://docs.fritz2.dev)
 [![Download](https://api.bintray.com/packages/jwstegemann/fritz2/fritz2-core/images/download.svg?version=0.4)](https://bintray.com/jwstegemann/fritz2/fritz2-core/0.4/link)
 
-fritz2 is a proof of concept for an ***extremely lightweight*** well-performing independent library for client-side ui in ***Kotlin*** heavily depending on coroutines and flows.
+fritz2 is a proof of concept for an ***extremely lightweight***, well-performing, independent library for client-side ui in ***Kotlin*** heavily depending on coroutines and flows.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,24 +29,24 @@ render {
 }.mount("target")
 ```
 
-fritz2 implements **precise data binding**. That means that exactly those (and **only** those) dom-nodes (elements, attributes, etc.) change, that depend on the parts of your data-model, that have changed. 
-There is no intermediate layer needed like a virtual DOM and you do not have to implement any additional methods to decide, which parts of your component have to be rerendered, when your data changes:
+fritz2 implements **precise data binding**. This means that when parts of your data model change, only the DOM-nodes depending on those parts will change as well.
```
jwstegemann, line 32: I think we need to emphasize this more, because this is one of the key benefits of fritz2 compared to react....
Those parts of the DOM (elements or attributes), exactly those and only those... 😇

README.md
```diff
@@ -58,18 +58,17 @@ You can either
 - server-communication (Rest APIs, etc.)
 - [documentation](https://docs.fritz2.dev) (work in progress)
 
-## What will come next?
+## What is next for fritz2?
 
 - performance and memory optimizations
```
jwstegemann, line 63: We should remove this list here and link to our GitHub issues...

README.md
EvaBS: I see what you mean, but that is pretty much already the meaning of "only the dom-nodes depending..". How about we change it to 
```suggestion
The key benefit when using fritz2 is its **precise data binding**. This means that when parts of your data model change, **only** the DOM-nodes depending on them will change as well (and no others).
```

README.md
EvaBS: Go ahead :-)

jwstegemann: just remove this part

README.md
jwstegemann: This means, that fritz2 automatically updates the DOM-nodes, all the DOM-nodes and nothing but the DOM-nodes depending on parts of your model that change.

PR #99, opened 2020-06-26
changed RootStore's initialData to Flow

### LINKED ISSUE #98: Second constructor in RootStore for Flows
With a second constructor in `RootStore<T>` you can use the data from an other store as `initialData` in your new store like:

```kotlin
val a = object: RootStore<String>("a") {}

val len: Flow<Int> = a.data.map { it.length }

val b = object: RootStore<Int>(len) {}
```

jamowei:
closes #98


jamowei:
This feature brings a lot of consequential problems. So it is still in progress...

jamowei:
We introduce the `action()` methods as an better alternative for this. Also this feature caused a lot of trouble.

PR #102, opened 2020-06-29
Fix mounting order

### LINKED ISSUE #101: Bug with rendering order of dynamic and static nodes
The following code
```kotlin
fun main() {
    val text = flowOf("test")

    render {
        div {
            text.map {
                render {
                    div {
                        text(it)
                    }
                }
            }.bind()
            div {
                text("hallo")
            }
        }
    }.mount("target")
}
```
results in
image

But that is a wrong order of resulting html elements.

jamowei:
fixes #101


jamowei:
added some (currently) failing test which shows the problem

Issue #92, opened 2020-06-21
Open tag classes


hpehl:
This is a feature request to open the element classes like A, Li, Div et al for inheritance. 

I'm developing an UI library based on [PatternFly](https://www.patternfly.org/v4/). In order to integrate the PatternFly tags into the fritz2 DSL, I'd like to extend from the existing HTML tags like `Li` or `Div`.

jamowei:
Hi, can you give us an concrete example for which you must extend the base HTML classes? Do you want to append additional attributes? This example would be great to make your intention clearer ;)

Did you already tried to open the HTML classes by your own? If yes did something din't work anymore?

hpehl:
Normally I don't want to add new attributes, instead I would like to re-use the reactive attributes defined in the various HTML classes. For instance I implement a class for a button in [PatternFly](https://www.patternfly.org/v4/documentation/core/components/button) which roughly looks like that:

```kotlin
class Button(
    modifier: Set<String>,
    private val text: String? = null,
    private val iconClass: String? = null,
    private val iconRight: Boolean = false
) : PatternFlyTag<HTMLButtonElement>(ComponentType.Button, "button", "button".component()),
    WithText<HTMLButtonElement> {
    ...
}
```

and 

```kotlin
open class PatternFlyTag<out E : HTMLElement>(
    componentType: ComponentType,
    tagName: String,
    baseClass: String
) : Tag<E>(tagName, baseClass = baseClass) {
    ...
}
```

Instead of using `dev.fritz2.dom.Tag` as base class, I'd like to extend from `dev.fritz2.dom.html.Button`. That would give me all the reactive attributes like 

```kotlin
    var autofocus: Flow<Boolean> by AttributeDelegate
    var disabled: Flow<Boolean> by AttributeDelegate
    var name: Flow<String> by AttributeDelegate
    var type: Flow<String> by AttributeDelegate
    var value: Flow<String> by ValueAttributeDelegate
```

See https://github.com/hal/halos-console/blob/master/src/jsMain/kotlin/org/patternfly/Button.kt for the full source code.

jamowei:
Hi Harald, did you tried this approach already?
```kotlin
// my own button component
// the receiver type of content function is Button instead of HtmlElements
fun HtmlElements.myButton(content: Button.() -> Unit): Button { 
    return button("btn my-btn") {
        name = const("my-button")
        domNode.onclick = {
            window.alert("My button was clicked!")
        }
        content()
    }
}

[...]

PR #104, opened 2020-07-01
Open tag classes for inheritance

hpehl:
Addresses #92

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,26 +1,26 @@
 name: build
 
 on:
-  push:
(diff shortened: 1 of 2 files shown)


jwstegemann:
Can you activiate "Actions/permissions/Enable local and third party Actions for this repository" in your repositories settings? I found some articles, that suggest, that this is needed so the checks are run before merging the PR...

hpehl:
Just checked that "Enable local and third party Actions for this repository" is enable in my fork at https://github.com/hpehl/fritz2

PR #105, opened 2020-07-01
Add single mount point for boolean attributes

hpehl:
Adds a single mount point for boolean attributes (flags) which is flexible enough to fulfill the following requirements:

1. Add an attribute (with an empty value), when the flow emits `true`, remove the attribute, when the flow emits `false`:

    ```kotlin
    val flow: Flow<Boolean> = ...
    render {
        div {
            flow.bindAttr("hidden")
        }
    }
    ```

    will lead to 

    ```html
    <div hidden/>
    ```

    when `flow` emits `true` and 

    ```html
    <div/>
    ```

    when `flow` emits `false`.

1. Map the boolean value to an arbitrary string:

    ```kotlin
    val flow: Flow<Boolean> = ...
    render {
        a {
            flow.bindAttr("aria-expanded", removeWhenFalse = false) {
                it.toString()
            }
        }
    }
    ```

    will lead to 

    ```html
    <a aria-expanded="true"/>
    ```

    when `flow` emits `true` and 

    ```html
    <a aria-expanded="false"/>
    ```

    when `flow` emits `false`.

1. A combination of 1 and 2:

    ```kotlin
    val flow: Flow<Boolean> = ...
    render {
        li {
            flow.bindAttr("aria-current") { "page" }
        }
    }
    ```

    will lead to 

    ```html
    <li aria-current="page"/>
    ```

    when `flow` emits `true` and 

    ```html
    <li/>
    ```

    when `flow` emits `false`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,26 +1,26 @@
 name: build
 
 on:
-  push:
+  pull_request:
     branches:
-      - '**'
+      - master
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up JDK 12
-      uses: actions/setup-java@v1
-      with:
-        java-version: 12
-    - name: Build and run Unit Tests
-      run: ./gradlew allTests
-    - name: Archive test-results
-      if: always()
-      uses: actions/upload-artifact@v1.0.0
-      with:
-        name: Test-Results
-        path: core/build/reports/tests/allTests
-    - name: Build with Gradle
-      run: ./gradlew build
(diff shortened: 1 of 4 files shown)


hpehl:
I'm open to discuss whether we really need the value function. The second example can also be done by 

```kotlin
val flow: Flow<Boolean> = ...
render {
    a {
        flow.map { it.toString() }.bindAttr("aria-expanded)
    }
}
```

Then we could *always* remove the attribute when the flow emits `false` and just pass a string to be used, when the flow emits `true` (which defaults to "").

WDYT?

hpehl:
Updated the PR with the simplified bindings.

PR #109, opened 2020-07-02
renamed to handleAndOffer + added first test

### LINKED ISSUE #100: Write test and docs for handleAndEmit inter-store-communication
Add some tests and new documentation for connecting stores by using the `handleAndEmit` method.
The method you must call inside of the function is called `offer` so maybe it would be a better idea to name the method `handleAndOffer` instead.

jamowei:
closes #100


jamowei:
Added new docs for this feature in fritz2-docs project.

PR #112, opened 2020-07-06
added a new Store (MapStore)

DrRingeDing:


jamowei:
Thanks for your PR! Before we can merge it, we need some running tests!
And maybe you can give us an example use case as Issue which we can link to this PR for documentation ;)

jamowei:
Also you KDoc is not entirely complete, please document all parts of your code.

jwstegemann:
Not necessary anymore. Can be realized using SubStores and Lenses accordingly. See removed FormatStore #136

PR #117, opened 2020-07-10
added dispatch functions for actions

jwstegemann:


jwstegemann:
comments missing. Should it already be this? 

@jamowei Could you please try it?

PR #127, opened 2020-07-20
Extending remote extension

jamowei:
Added new append() function and added a new test which uses crudcrud.com


core/src/jsMain/kotlin/dev/fritz2/remote/remote.kt
```diff
@@ -97,60 +97,71 @@ open class Request(
     /**
      * issues a get request returning a flow of it's response
```
jwstegemann, line 98: Ok

PR #149, opened 2020-08-03
Refactor validation entirely

### LINKED ISSUE #148: Convert Validation extension to a service
For providing a clean and similar API in fritz2 it is needed to convert the existing validation extension to the new service design. So this features is then similar e.g. to the `EntityService` and other. Therefore a `RootStore` must not implement the `Validation` interface any more and the validation API gets cleaner.

jamowei:
Fixes #148

### DIFF
--- a/core/src/commonTest/kotlin/dev/fritz2/validation/validation.kt
+++ b/core/src/commonTest/kotlin/dev/fritz2/validation/validation.kt
@@ -1,0 +1,51 @@
+package dev.fritz2.validation
+
+import dev.fritz2.lenses.Lenses
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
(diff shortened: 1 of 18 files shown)


core/src/commonTest/kotlin/dev/fritz2/validation/validation.kt
```diff
@@ -0,0 +1,50 @@
...
+
+class Message(val text: String) : ValidationMessage {
+    override fun isError(): Boolean = true
+}
+
+val carNameIsBlank = Message("car name can not be blank")
+val colorValuesAreToLow = Message("color members are lower then 0")
+val colorValuesAreToHigh = Message("color members are greater then 255")
+
+val carValidator = object: Validator<Car, Message, Unit>() {
```
jwstegemann, line 18: Why not using factory-method here?

PR #168, opened 2020-09-04
#167: No trailing slash in remote

### LINKED ISSUE #167: URLs in remote always end with a '/'
**Describe the bug**
URLs build with the `remote()` function always have a trailing slash. 

**To Reproduce**
The following code 

```kotlin
val json = remote("https://randomuser.me/api/?results=3")
        .acceptJson()
        .get()
        .getBody()
```

results in an URL of https://randomuser.me/api/?results=3/ instead of https://randomuser.me/api/?results=3. This leads to errors in the backend service (`results` parameter is no longer recognized).

**Expected behavior**
URLs build with `remote()` should not automatically end with a slash.

hpehl:
Fixes #167


jwstegemann (approved):
Thank you for discovering the bug and providing the correction at once!

PR #172, opened 2020-09-10
#171: Fix boolean attribute delegates

### LINKED ISSUE #171: Fix boolean attribute delegates
**Describe the bug**
Many tags have boolean attributes like `button.disabled`. These attributes are currently handled by `AttributeDelegate`. This class sets the attribute value to `true` resp `false`:

```html
<button disabled="true">Click me!</button>
<button disabled="false">Click me!</button>
```

**Expected behavior**
Instead of setting the attribute value, the attribute itself should be added / removed for boolean attributes:

```html
<button>Click me!</button>
<button disabled>Click me!</button>
```

hpehl:
Fixes #171 

This PR modifies all boolean attributed in `core/src/jsMain/kotlin/dev/fritz2/dom/html/elements.kt`. This assumes that all boolean attributes work the same way: When `true` the attribute is set, when `false` the attribute is removed.

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/dom/attributes.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/dom/attributes.kt
@@ -16,6 +16,15 @@ internal object AttributeDelegate {
     }
 }
 
+internal object BooleanAttributeDelegate {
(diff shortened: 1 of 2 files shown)


jamowei (approved):
Thanks!

PR #175, opened 2020-09-23
Websocket support

### LINKED ISSUE #116: Support web-sockets in remote
I would like to communicate with a backend that uses websockets, and flows seem to lend themselves nicely to that.

```kotlin
val socket = remote("ws://server/ws").websocket() {
  // transform frames here
} handledBy update

// In other places:
store.data.bind() // Render incoming data
socket.send(message) // Send a message
```

I am still unfamiliar with fritz2, but as an alternative I think that I could use `window.fetch` and bind to the events.

jamowei:
fixes #116


core/src/jsMain/kotlin/dev/fritz2/binding/store.kt
```diff
@@ -150,6 +167,20 @@ inline fun <T, R> Store<T>.syncBy(handler: Handler<R>, crossinline mapper: suspe
...
+    session.messages.body.map {
+        val received = resource.serializer.readList(it)
+        last = received
+        received
+    } handledBy update
+
+    data.drop(1).onEach {
+        if (last != it) session.send(resource.serializer.writeList(it))
+    }.watch()
+}
```
jwstegemann, line 182: Have to keep this in mind when reworking coroutine-scopes

jwstegemann (approved):
Great!

PR #176, opened 2020-09-25
Styling and Components

jwstegemann:
Offer a way to

- write css in Kotlin
- create dynamic css
- parametrize components with styles like styled-system.com

Add a base set of components


jwstegemann:
First demo in subproject components (app.kt) (will not stay there)

PR #178, opened 2020-10-02
Annotator processor visibility option

mpetuska:
kapt option 'fritz2.visibility' to specify lens object visibility and default to explicit 'public' to allow using it for library authors with explicitApi() flag

Specify in consumer projects as
```kotlin
kapt {
  arguments {
    arg("fritz2.visibility", "internal")
  }
}
```


mpetuska:
@jwstegemann ?

jwstegemann:
At first: Sorry for getting back to you on this late!  To be honest, Jan and I were both busy enjoying our outdoor-hobbies using the (probably) last days of good weather over here...

Thank you very much for your contribution. `explicitApi` was not on our roadmap but since fritz2 is intended as delivering just the basics it is of course a good idea.

jwstegemann (approved):

PR #189, opened 2020-11-19
moved kitchenSink to sub project

jwstegemann:


chausknecht (approved):
- checked components still builds without problems
- checked that new kitchenSink project builds and runs without problems

PR #194, opened 2020-11-23
Dokka upgrade

jamowei:
- Upgraded version of Dokka to newest version
- added multiproject Dokka task for core, styling & components


chausknecht:
It still builds and everything gets into /api.
But I don't see any visible changes within the HTML documentation itself. The customization seems not to wqork yet.

jamowei:
The changes will only be available after we pulished the new fritz2 version 0.8.

chausknecht:
As the main testable part works (building) , I think we should close this :-)

chausknecht:
Looks good to me!

chausknecht (dismissed):

chausknecht (approved):

PR #195, opened 2020-11-24
a lot of cleanup-work on styling, theme and components. 

jwstegemann:
Mostly to access the current theme by receiver not explicitly.

Do not like the way inputFields have to be used now... Some Brainstorming necessary.


components/src/jsMain/kotlin/dev/fritz2/components/formcontrol.kt
```diff
@@ -283,7 +285,7 @@ open class FormControlComponent {
         renderContext.div {
             helperText?.let {
                 (::p.styled {
-                    color { theme().colors.dark }
+                    color { Theme().colors.dark }
```
chausknecht, line 288: just use ``dark``

components/src/jsMain/kotlin/dev/fritz2/components/formcontrol.kt
```diff
@@ -296,7 +298,7 @@ open class FormControlComponent {
             errorMessage.render {
                 if (it.isNotEmpty()) {
                     lineUp({
-                        color { theme().colors.danger }
+                        color { Theme().colors.danger }
```
chausknecht, line 301: just use ``danger``

components/src/jsMain/kotlin/dev/fritz2/components/formcontrol.kt
```diff
@@ -351,7 +353,7 @@ class SingleControlRenderer(private val component: FormControlComponent) : Contr
             id = id,
             prefix = prefix
         ) {
-            spacing { theme().space.tiny }
+            spacing { Theme().space.tiny }
```
chausknecht, line 356: just use ``tiny``

kitchenSink/src/jsMain/kotlin/flexBoxDemo.kt
```diff
@@ -68,7 +70,7 @@ fun RenderContext.flexBoxDemo(themeStore: ThemeStore, themes: List<ExtendedTheme
                     left { "0" }
                 }
             }) {
-                themeStore.data.map { currentThemeIndex ->
+                ThemeStore.data.map { currentThemeIndex ->
```
chausknecht, line 73: Is it really better style to rely on a global variable here?
Passing the store instead of the Handler is imho clearer, as now one dependency is only hidden and the user has to cope with an extra Flow that otherwise was easier to identify!

kitchenSink/src/jsMain/kotlin/inputDemo.kt
```diff
@@ -139,7 +162,7 @@ fun RenderContext.inputDemo(): Div {
                 baseClass: StyleClass? = null,
                 id: String? = null,
                 prefix: String = "ourInputField",
-                init: Input.() -> Unit
+                init: InputFieldComponent.() -> Unit
```
chausknecht, line 165: Should be renamed to ``build``!

chausknecht (commented):
I will directly change the comments I made!

kitchenSink/src/jsMain/kotlin/flexBoxDemo.kt
chausknecht: As time is short, I consider thsi a minor issue to discuss.
So no changes for now ;-)

chausknecht:
I have resolved all remarks

chausknecht (approved):

PR #198, opened 2020-11-26
added wide sizes

jwstegemann:
... to styling dsl


chausknecht (approved):
- nice and fitting solution!
- added some api doc in order to keep the idea
- adopted some predefined styles of modal dialogs

PR #206, opened 2020-12-02
updated input demo

EvaBS:
Updated page to agreed formats.

Other notes:
I used the simpleLinkWithBackground from app.kt for midtext links to other components. That style should be re-worked by Moritz.
Added notes about form control (using labels and validation). 
Added a few input type examples.


chausknecht (approved):
Looks good!

PR #223, opened 2020-12-07
Changes to colors, theme values, styling

jwstegemann:


jwstegemann:
right now button-colors are not rendered because `primary` is hardcoded in `DefaultTheme`. This has to respect `main-color` css-property set by `color`-property of button component. This color has also to be respected by hover.

PR #230, opened 2020-12-10
Remove RenderContext as Receiver

### LINKED ISSUE #227: Dynamic content within our stack components does not work
**Describe the bug**
If you want to show some content based upon a flow and use a ``stackUp`` or ``lineUp`` as container, the items are rendered at the wrong dom node level (above the parent, not as children!)

**To Reproduce**
Just execute the following snippet:

        div {
            val store = storeOf(listOf("A", "B", "C"))
            stackUp {
                items {  // this == RenderContext, but for the rendering ``Div`` is used
                    p { +"Should appear before items" }
                    store.renderEach({ it }) {
                        box { it.data.asText() }
                    }
                    p { +"Should appear after items" }
                }
            }
        }

If you run this code, the three items from the store are rendered as direct childrin of the enclosing ``div``!
That will looke like this:

    A
    B
    C
    Should appear before items
    Should appear after items

**Expected behavior**
The items should be rendered at the correct position of the dom tree!
This would lead to this output:

    Should appear before items
    A
    B
    C
    Should appear after items

**Additional context**
Obviously the receiver type of the component's ``item`` property and the *used* type by the component factory function does not match: ``RenderContext`` vs ``Div``.
 - change receiver of ``items`` to ``Div``
 - scan whole components library for similar cases (places, where unspecific ``RenderContext`` is used)

chausknecht:
- fixes #227
- remove ``RenderContext`` as receiver
- add specific Tag as receiver or ``Tag<HtmlElement>`` for _generic_ cases

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/button.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/button.kt
@@ -4,6 +4,8 @@ import dev.fritz2.dom.Listener
 import dev.fritz2.dom.WithEvents
 import dev.fritz2.dom.html.Button
 import dev.fritz2.dom.html.RenderContext
+import dev.fritz2.dom.html.Span
+import dev.fritz2.dom.html.TextElement
 import dev.fritz2.styling.StyleClass
 import dev.fritz2.styling.StyleClass.Companion.plus
 import dev.fritz2.styling.params.*
@@ -168,7 +170,7 @@ open class PushButtonComponent {
         label = { hide -> span(if (hide) hidden.name else null) { value.asText() } }
     }
 
-    var loadingText: (RenderContext.() -> Unit)? = null
+    var loadingText: (Button.() -> Unit)? = null
 
     fun loadingText(value: String) {
         loadingText = { span { +value } }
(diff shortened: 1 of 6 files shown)


chausknecht:
Here is some runnable code for some external project in order to test most changes:

    render {
        val aList = listOf("A", "B", "C")
        val bList = listOf("D", "E", "F")
        val store = storeOf(aList)
        val triggerz = storeOf(Theme().icons.fritz2)

        div {
            clickButton { text("Open") } handledBy modal {
                content {
                    store.renderEach({ it }) {
                        box { it.data.asText() }
                    }
                }
            }
        }

        div {
            h3 { +"Popover" }
            popover {
                trigger {
                    triggerz.data.render { icon { def = it } }
                }
                header { store.data.asText() }
                content { store.data.asText() }
                footer { store.data.asText() }
                placement { right }
            }
        }

        div {

            clickButton { text("toggle") }.map {
                if (store.current[0] == "A") bList else aList
            } handledBy store.update

            clickButton {
                text(store.data.asString())
                loading(flowOf(true))
                loadingText(store.data.asString())
            }
        }

        div {
            stackUp {
                items {
                    p { +"Should appear before items" }
                    lineUp {
                        items {
                            store.renderEach({ it }) {
                                box { it.data.asText() }
                            }
                        }
                    }
                    p { +"Should appear after items" }
                }
            }
        }

    }.mount("target")

jamowei (approved):

PR #231, opened 2020-12-11
Optimize patch handling

chausknecht:
- patches within Myer's diff algorithm are now delivered as ``List`` and not as ``Flow`` anymore.
- obviously the browser's rendering engine are optimized to apply bigger DOM changes at once, so the overall rendering speed is much faster now.

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/binding/mount.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/binding/mount.kt
@@ -16,7 +16,7 @@ import kotlinx.coroutines.flow.scan
  */
 //TODO: inline?
 fun <T> mountSingle(parentJob: Job, upstream: Flow<T>, set: suspend (T, T?) -> Unit) {
-    (MainScope() + parentJob).launch {
+    (MainScope() + parentJob).launch(start = CoroutineStart.UNDISPATCHED) {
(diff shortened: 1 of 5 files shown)


chausknecht:
In order to test the efforts here is some demo code, that should be run before applying this patch and afterwards:

    val a = generateSequence { Random.nextInt(0, 1000) }.take(1000).toList() // change values to test different sizes
    val b = generateSequence { Random.nextInt(0, 1000) }.take(500).toList()
    val row = generateSequence { Random.nextInt(0, 1000) }.take(50).toList()

    val toggle = storeOf(a)
    val rowStore = storeOf(row)

    div {
        clickButton {
            text(toggle.data.map { it.size.toString() })
        } handledBy toggle.handle { old -> if (old.size == b.size) a else b }
        toggle.data.renderEach {
            (::p.styled {
                background { color { if (it % 2 == 0) danger else warning } }
                fontSize { "${it % 10}px" }
                color { "white" }
            }) {
                rowStore.current.forEach {
                    span { +it.toString() }
                }
            }
        }
    }

Just use the _toggle_ button in order to switch between different sized datasets.

With applied patch the change with this configuration is less than one second on my computer.
Tested with 
- firefox 83.0
- chrome 87.0.4280.88

jamowei (approved):
No issues detected. Nice work!

PR #232, opened 2020-12-14
Refactor repositories API

jamowei:
Improve API for Repositories


jwstegemann:
Would it be a good idea to separate RessourceSerializer and RessourceListSerializer so you don't have to implement methods zog won't need?

jwstegemann:
@jamowei what do you think?

jamowei:
> @jamowei what do you think?

You are right, I changed the default implementation if `RessourceSerializer` and did a bunch of other things. Please review again.

jwstegemann (approved):

PR #235, opened 2020-12-17
cnconsult neumann/toast

cnconsult-neumann:


jwstegemann:
Several points here:

- shadows are not working correctly (leaves white frame, etc.)
- close button does not work
- `standAlone` breaks api. Why not offer two factories like `showToast` and `toast`
- offer convenience-factories like `toast.info`, `toast.error`, etc. This would make `customComponent` unnecessary, when the base `toast` is just empty and has `content`....
-

haukesomm:
> - shadows are not working correctly (leaves white frame, etc.)
> - close button does not work

@jwstegemann I've addressed the points regarding the `standAlone` attribute and convenience factories but haven't been able to reproduce the cited issues on either Chrome, Firefox or Safari unfortunately. Both the close-buttons and shadows are working fine on my machine.

Do you have some more information on the issue? Maybe I'm also just missing something... 🤔

jwstegemann:
> > * shadows are not working correctly (leaves white frame, etc.)
> > * close button does not work
> 
> @jwstegemann I've addressed the points regarding the `standAlone` attribute and convenience factories but haven't been able to reproduce the cited issues on either Chrome, Firefox or Safari unfortunately. Both the close-buttons and shadows are working fine on my machine.
> 
> Do you have some more information on the issue? Maybe I'm also just missing something... 🤔

I tested with the toasts-branch in kitchensink-repo. After your improvements this does not work anymore. Do you have some demo / test available?

haukesomm:
I tested with a copy of the fritz2 project template as I didn't know there was a branch in kitchensink already 😅 I'll make it available to you tomorrow so you can have a look as well.

haukesomm:
While the alerts are added to kitchenskink, an example of the different alert variants and styles can be found in [this](https://github.com/haukesomm/fritz2-playground) testing repo.

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
```diff
@@ -0,0 +1,280 @@
...
+ * If no severity is specified, 'info' will be used by default.
+ *
+ * Additionally, a number of different layout options are available. These are:
+ * - 'subtle': A subtle style using different shades of the severity's base color defined in the application theme.
+ * - 'solid': A solid style using the severity's color from the application theme and a solid white color for the icon,
+ * text and decorations.
+ * - 'Top-Accent': A variation of the 'subtle' variant with a decoration element at the top.
+ * - 'Left-Accent': A variation of the 'subtle' variant with a decoration element on the left.
+ * If no variant is specified, 'solid' is used by default.
+ *
```
EvaBS, line 34: 1) From the explanation in the kitchensink demo and from what I gather from these comments, I the "subtle" style is supposed to be default. However, if no variant is specified, the "solid" style appears. I assume that the demo text is correct and the default is incorrectly set in the component.

2)

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
```diff
@@ -0,0 +1,280 @@
...
+ * Different styles based on severities are supported, as well as a number of different layout options.
+ *
+ * Currently the following severities are available:
+ * - Info
+ * - Success
+ * - Warning
+ * - Error
+ * Specifying a severity will change the alert's color scheme based on the colors defined in the application theme.
+ * If no severity is specified, 'info' will be used by default.
+ *
```
EvaBS, line 26: Is the background color the only thing that changes on specifying a severity? If so, shouldn't this be a property with context and accept all ColorProperties (and a different name, backgroundColor maybe)?

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
```diff
@@ -0,0 +1,280 @@
...
+ *     variant { leftAccent }
+ * }
+ * ```
+ */
+@ComponentMarker
+class AlertComponent {
+
+    companion object {
+        private const val accentDecorationThickness = "4px"
+    }
```
EvaBS, line 56: Does this component need size options, and with it a changing thickness value?

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
```diff
@@ -0,0 +1,280 @@
...
+
+/**
+ * Convenience extension to display a [ComponentValidationMessage] as an alert.
+ * The alert's severity and content are determined from the validation message's properties.
+ *
+ * @param renderContext RenderContext to render the alert in.
+ */
+fun ComponentValidationMessage.asAlert(renderContext: RenderContext) {
+    renderContext.alert {
+        severity { when(severity) {
```
EvaBS, line 273: Currently, Severity cannot be Success

styling/src/jsMain/kotlin/dev.fritz2.styling/theme/default.kt
```diff
@@ -2286,6 +2292,174 @@ open class DefaultTheme : Theme {
...
+                        css("visibility: hidden")
+                    }
+                    override val decorationTop: Style<BasicParams> = {
+                        background { color { it } }
+                    }
+                }
+            }
+        }
+    }
+
```
EvaBS, line 2388: I would love to have the bottom and rightAccent variants as well (might not be called variant anymore, tho)

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
chausknecht: This is due to the fact, that ``Severity`` currently does not have a ``Success`` value. I have allready added this within my enhancements of formControl (PR#256), but that should be reflected here as well of course. So in order to be compliant here is the code from my branch that should be applied here as well:
```kotlin
enum class Severity {
    Info, Success, Warning, Error
}
```
(The order matters! So ``Success`` should be in second position!)

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
chausknecht: No, afaik the opposite is true: The comment and the demo should be changed though.
But Jens is the source of truth for this :-)

EvaBS: The leftAccented and topAccented variants are all colored "subtle".

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
EvaBS: Exactly my point, good to know it's already in progress.

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
EvaBS: You are right, Jens had already changed that, so the demo was off because he didn't change it there. Sorry!

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
haukesomm: Actually @jwstegemann pointed out that the icon should change as well, so I'm not sure whether I can get rid of the context. If it only was the background color that changes (as I initially though too) I would agree.

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
haukesomm: At least there is no such option in Chakra UI, I believe. It would be very easy to add the option, though. What do you think?

jwstegemann (approved):
Looks pretty good now. Will open a separate issue to set icon on alerts depending on severity...

PR #237, opened 2020-12-22
Buttons: Disabling via base context. TextArea: Unified disabled/readOnly.

EvaBS:
Same changes were applied to kitchensink branch.


jwstegemann:
will be dealt with when applying new component guidelines to all existing components....

PR #244, opened 2021-01-19
Components validation

jamowei:
If added a first draft of component validation. Please review!
Take a look [here](https://gist.github.com/jamowei/d485f7a1c588ffa38929979d6bade5e4) for running example.


chausknecht:
Looks quite nice to me!

I believe you prefered to remain flexible with the type of the store by relying on runtime type checking within the ``fun <D> Store<D>.validationMessage(): Flow<String>`` function instead of defining some ``ValidateableStore``, which all of our components then would have to adopt, even if there is no need of validation? (A possible typed variant would require some kind of null pattern for the validator in order to remain kinda *optional* though.)

I would like to tinker a small example for our ``formControl`` component in order to get a *feeling* how this approach fits there. My initial guess is, that this would work very good, as the store is injected into the underlying sub component, but the validation is handled by the formControl itself. With this kinda dynamic approach one could *try* to validate but catch the ``RootStoreIsNotValidatable`` if validation is not possible. Thus the user is not forced to implement validation if not needed.

components/src/jsMain/kotlin/dev/fritz2/components/validation/validation.kt
```diff
@@ -0,0 +1,38 @@
...
+            if (validator.isValid(new, metadata)) new else old
+        })
+    }
+}
+
+class RootStoreIsNotValidatable(id: String) :
+    Exception("RootStore of data with id=$id must implement Validatable interface")
+
+fun <D> Store<D>.validationMessage(): Flow<String> = when (this) {
+    is RootStore<*> -> {
```
jwstegemann, line 24: You could declare this method here abstract and implement it at RootStore and SubStore. This would fit better to all the other methods that have different behavior for Root and Sub.

components/src/jsMain/kotlin/dev/fritz2/components/validation/validation.kt
```diff
@@ -0,0 +1,38 @@
...
+        if (this is Validatable<*, *>) {
+            this.validator.find { it.id == this@validationMessage.id }.map { it?.message ?: "" }
```
jwstegemann, line 26: What about returning all messages for a given id here and leave it to the user to take just the first if needed? Or offer both methods? (validationMessage and validationMessages)

components/src/commonMain/kotlin/dev/fritz2/components/validation/validation.kt
```diff
@@ -0,0 +1,23 @@
+package dev.fritz2.components.validation
+
+import dev.fritz2.validation.ValidationMessage
+import dev.fritz2.validation.Validator
+
+data class ComponentValidationMessage(val id: String, val severity: Severity, val message: String) : ValidationMessage {
```
jwstegemann, line 6: I think we should add a title: String. Maybe optional. This is often needed for toasts, etc.

components/src/jsMain/kotlin/dev/fritz2/components/validation/validation.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import dev.fritz2.binding.RootStore
+import dev.fritz2.binding.Store
+import dev.fritz2.binding.SubStore
+import dev.fritz2.validation.Validator
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.map
+
+interface Validatable<D, T> : Store<D> {
```
jwstegemann, line 11: I think the name Validatable might be misleading here, because the point is if it really gets validated in this store, not if it is possible to validate. What do you think about HasValidation or WithValidator or just Validating

components/src/jsMain/kotlin/dev/fritz2/components/validation/validation.kt
jamowei: I don't want to change our core implementation for a special case in components. In my opinion this extension method approach fits the needs and provides a better separation of the different code bases.

components/src/jsMain/kotlin/dev/fritz2/components/validation/validation.kt
jamowei: Yes, good point. I change the returning type of `validationMessage` function to `Flow<ComponentValidationMessage?>` which more general and can be mapped afterwards. Additionally I have added a second extension method `validationMessages` which return a `Flow<List<ComponentValidationMessage>>` in case you have more then one message for field in your data model.

components/src/jsMain/kotlin/dev/fritz2/components/validation/validation.kt
jamowei: Ok, renamed to `WithValidator` like we had in our core code.

components/src/commonMain/kotlin/dev/fritz2/components/validation/validation.kt
jamowei: Good point! I come up with the following now:
```kotlin
data class ComponentValidationMessage(
    val id: String,
    val severity: Severity,
    val message: String,
    val details: String? = null,
)
```
Hopefully this fits the most of possible use cases!

jamowei:
Ok, I have done the rework. Please review again and if there is noting left to do, I will start documenting.
@jwstegemann @chausknecht

jwstegemann (approved):

PR #245, opened 2021-01-19
Update LICENSE file

jamowei:
Set year to 2021.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Jens Stegemann
+Copyright (c) 2021 Jens Stegemann
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


chausknecht:
How about other license files we include and deliver?
``styling\src\jsMain\kotlin\dev.fritz2.styling\theme\icons\LICENSE.md``

Is it relevant to update thoese files or might it even be better not to change these things and to preserve the status quo of the insertion timestamp?

jwstegemann:
When I understand the legal aspect correctly we should update the copyright year whenever changes are made.
So we should upgrade our own license (since we already made changes in 2021), but should not update the icons licenses right now.

jwstegemann (approved):

jwstegemann:
Shouldn't we change this to "by Jens Stegemann & Jan Weidenhaupt"? Just for the case...

PR #249, opened 2021-01-20
Fix placement of modals

### LINKED ISSUE #229: Change default HTML Template in order to make modals work
**Describe the bug**
Currently the default HTML Template defines the mount id on the ``body`` tag.
This conflicts with the modal component, because the content of the body tag is replaced by the initial ``mount`` and so the modals div gets removed.

**Suggestion**
- Insert ``div`` as mount target

jamowei:
Fixes #229


jamowei:
See my test code [here](https://gist.github.com/jamowei/6b88ce9b8bb6a29731d36eb40886d81e)

chausknecht (approved):
Looks good!

I have tested this with our KitchenSink project and everything works like a charm.

PR #250, opened 2021-01-21
harmonizeComponentsApi

chausknecht:
- Apply [components guidlines](https://github.com/jwstegemann/fritz2/wiki/Interne-Komponentenrichtlinien) to all existing fritz2 components
- reduce implementation complexity by
  - using specific properties abstractions
  - using "traits" for common property groups


chausknecht:
Some trouble with branches so this PR will be reopened elsewhere!

PR #251, opened 2021-01-22
Added formControl support for: selectField, switch, textArea

EvaBS:
+ Minor changes.


chausknecht (changes requested):
Looks very good!

One thing should be improved before merging: A radioGroup is also generic, so the factoy method of formControl should reflect this too as for checkboxGroup.

chausknecht (approved):

PR #255, opened 2021-01-29
shadesOfGray

mkempa-np:
This branch introduces 6 shades of gray to the theme colors, replacing light and lightEffect with lightGray and lighterGray and adding 4 more shades of gray.


chausknecht (approved):
Very nice shades of gray!

PR #258, opened 2021-02-05
File Selectors

jamowei:
I have added the first draft of a file selector button.
I have also added a PR for the new demo page in [fritz2-kitchensink](https://github.com/jwstegemann/fritz2-kitchensink/pull/14) for testing the component.


components/src/jsMain/kotlin/dev/fritz2/components/file.kt
```diff
@@ -0,0 +1,252 @@
...
+    build: FileSelectionComponent.() -> Unit = {}
+): Flow<File> {
+    var file: Flow<File>? = null
+    val component = FileSelectionComponent().apply(build)
+    (::div.styled(styling, baseClass, id, prefix) {}) {
+        val inputElement = (::input.styled { display { none } }) {
+            type("file")
+            component.accept?.invoke(this)
+            file = changes.events.mapNotNull {
+                domNode.files?.item(0)
```
chausknecht, line 178: Might still let file to be null right? 
If so, the return in line 186 could fail!

Could this also be done without a nullable var? Or with a ``Flow<File?>`` instead?

components/src/jsMain/kotlin/dev/fritz2/components/file.kt
```diff
@@ -0,0 +1,252 @@
...
+ * @return a [Flow] that offers the selected [File]
+ */
+fun RenderContext.files(
+    styling: BasicParams.() -> Unit = {},
+    baseClass: StyleClass? = null,
+    id: String? = null,
+    prefix: String = "file",
+    build: FileSelectionComponent.() -> Unit = {}
+): Flow<List<File>> {
+    var files: Flow<List<File>>? = null
```
chausknecht, line 226: Same as above: Could this be solved without null? For collections often an empty instance works.

chausknecht (changes requested):

chausknecht (approved):

PR #262, opened 2021-02-10
added app-component (pwa) and nav-components

jwstegemann:
find working example at https://github.com/jwstegemann/fritz2-app
click-handling for navLinks not implemented yet
documentation still missing...


components/src/jsMain/kotlin/dev/fritz2/components/appFrame.kt
```diff
@@ -0,0 +1,281 @@
...
+ * - nav
+ * - main
+ * - footer (only rendered if defined)
+ * - tabs (only rendered if defined)
+ *
+ * In addition to that you can define how the sidebarToggle on small screens is rendered.
+ * By default is a hamburger-button.
+ *
+ * The rendering function is used by the component factory functions [appFrame], so it is
+ * not meant to be called directly unless you plan to implement your own button.
```
chausknecht, line 35: Do you mean your own app frame component? (Not Button!)

components/src/jsMain/kotlin/dev/fritz2/components/appFrame.kt
```diff
@@ -0,0 +1,281 @@
...
+            margins { left { "-.5rem" } }
+        }) {
+            variant { link }
+            icon { fromTheme { menu } }
+        } handledBy sidebarToggle
+    }
+    var nav = ComponentProperty<TextElement.() -> Unit> {}
+    var main = ComponentProperty<TextElement.() -> Unit> {}
+    var footer = ComponentProperty<(TextElement.() -> Unit)?>(null)
+    var tabs = ComponentProperty<(Div.() -> Unit)?>(null)
```
chausknecht, line 119: all Properties should be ``val`` instead of ``var``

components/src/jsMain/kotlin/dev/fritz2/components/nav.kt
```diff
@@ -0,0 +1,79 @@
...
+                left {
+                    color { lightGray.important }
+                }
+            }
+        }
+    }
+
+    var icon = ComponentProperty<IconComponent.() -> Unit> { fromTheme { bookmark } }
+    var text = DynamicComponentProperty<String>(flowOf("Navigation Link"))
+    var active = ComponentProperty<Flow<Boolean>?>(null)
```
chausknecht, line 34: All properties should be ``val`` instead of ``var``

styling/src/jsMain/kotlin/dev.fritz2.styling/theme/definitions.kt
```diff
@@ -663,3 +660,22 @@ interface ToastStatus {
...
+    val sidebar: Style<BasicParams>
+    val nav: Style<BasicParams>
+    val footer: Style<BasicParams>
+    val header: Style<FlexParams>
+    val main: Style<BasicParams>
+    val tabs: Style<FlexParams>
+    val navLink: Style<FlexParams>
+    val navSection: Style<BasicParams>
+}
+
```
chausknecht, line 681: How about separating the navLink component styles from the appFrame by extracting a separate interface within the theme and just refering to it by the ``AppFrameStyles``? (This has allready been used for ``SeverityAware`` as most forms are styled in exactly the same way)

This approach has the following benefits:
- does not mix different components and their stylings
- does not close the possibility to separate the styling between appFrame and content navigation - even though this seems to be an unusual use case

The added complexity is rather small imho.

chausknecht (changes requested):

styling/src/jsMain/kotlin/dev.fritz2.styling/theme/definitions.kt
jwstegemann: navLink and navSection should not be used outside off appFrame because of their responsive behavior that does not make sense elsewhere. added this to the docs

chausknecht (approved):

PR #263, opened 2021-02-12
Improve API of formControl

chausknecht:
- removed required property
- open formControl to be extended by new wrapping functions and renderers
- correct some typo in basic abstraction


derTranspirator (approved):
looks good

PR #265, opened 2021-02-15
authenticated request

petertiupe:
authenticated request incl. tests added. Class ist Authenticated Request


jamowei (approved):
Good work!

PR #270, opened 2021-02-20
support values and valuesAsNumber on input events as well

jillesvangurp:
I ran into this when trying to capture the input events that the input slider fires ias you drag it. It fires a change event when you stop dragging. Likewise, useful on text input to capture values as you type.

I modified the test to test both events.

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/dom/listener.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/dom/listener.kt
@@ -5,6 +5,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 import org.w3c.dom.*
 import org.w3c.dom.events.Event
+import org.w3c.dom.events.InputEvent
 import org.w3c.dom.events.KeyboardEvent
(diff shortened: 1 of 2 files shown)


jillesvangurp:
I encountered some flakiness with the tests. I fixed it for the test I modified but probably would be good to generalize the approach and fix all tests. I've had various other tests fail and not fail on repeated gradle builds.

jillesvangurp:
I could use some help with this test. It seems to still fail in CI/gradle. Works fine in intellij. The number of tests that fails is different every time I run this so it looks like there are bigger underlying issues that need fixing. 

FYI, I can't get chrome / chrome headless to work with  with Karma so I configured it locally to use Firefox.

jamowei:
@jillesvangurp Thank you for the PR. I will look into it. Can you give me access to your repo then I can do some changes there and hopefully we can merge your PR soon :)

jillesvangurp:
Done, I gave you access.

jamowei (approved):
I have cleaned up the test a bit. Now it is running with both events registered on same input element :)

PR #273, opened 2021-02-22
Fix AppFrame height on Firefox

haukesomm:
This PR fixes the AppFrame not expanding to full vertical height on Firefox.
This behavior affects Firefox on all operating systems.

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/appFrame.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/appFrame.kt
@@ -41,7 +41,7 @@ open class AppFrameComponent() {
             staticStyle(
                 """
                 body {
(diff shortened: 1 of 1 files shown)


jamowei:
Works now in my Firefox and Chrome. Don't know if this breaks other devices...?

haukesomm:
> Don't know if this breaks other devices...?

As I can tell from testing with BrowserStack it does not break. But as you're mentioning it, maybe this should be tested with some more complex project using the AppFrame.

haukesomm:
@jamowei I did some more in-depth testing and can confirm the fix doen't affect the AppFrame on any common browser/os combination.

jamowei (approved):
Thanks for testing!

PR #277, opened 2021-02-25
Window Object, Listener rework, Popover features

derTranspirator:
- new Window object
- WindowListener, DomListener and Listener
- Window Test (2 tests set on ignore because Karma can't handle the JS Methods which we needs into the tests)

- Popover Component get two new feature
-- closeOnBlur
-- closeOnEscape


jamowei (approved):
Nice work!

PR #293, opened 2021-03-05
Migration to mavenCentral

jwstegemann:


jwstegemann:
processes for snapshots and releases are working from local pc. we have to test publishing from github actions, when this pr gets merged...

docs (for snapshots) have to be adjusted, when this pr is merged because all subsequent snapshot-builds will be published to sonatype instead of bintray

@hpehl Thank you for the tipp, worked much better than I expected ;-)

jwstegemann:
created to PRs at docs and template

jamowei (approved):

Issue #294, opened 2021-03-05
Move styling of inputField component into the theme


chausknecht:
In order to ease the custom styling within a custom theme, all of the basic styling ``basicInputStyles`` should be moved into the styling of ``inputField`` within our default theme.

Whoever fixes this bug should carefully inspect all other components and look for similar artifacts that should also be refactored.

This should not change the API of the component!

mkempa-np:
The call of basicInputStyles in the method render() overrides the styling of the variant and size of the inputField. This shoud not happen. When moving the basic styling to the theme, the call of basicInputStyles should be erased from the render method.

PR #298, opened 2021-03-05
moves basicInputStyles of inputFieldComponent to default theme.

mkempa-np:
fixes issue #294

### DIFF
--- a/styling/src/jsMain/kotlin/dev.fritz2.styling/theme/default.kt
+++ b/styling/src/jsMain/kotlin/dev.fritz2.styling/theme/default.kt
@@ -1320,15 +1320,14 @@ open class DefaultTheme : Theme {
         override val sizes = object : FormSizes {
             override val small: Style<BasicParams> = {
                 height { "2rem" }
-                minWidth { "2.5rem" }
                 fontSize { small }
                 paddings {
                     horizontal { tiny }
                 }
             }
+
             override val normal: Style<BasicParams> = {
                 height { "2.5rem" }
-                minWidth { "2.5rem" }
                 fontSize { normal }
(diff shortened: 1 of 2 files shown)


styling/src/jsMain/kotlin/dev.fritz2.styling/theme/default.kt
```diff
@@ -1346,12 +1346,54 @@ open class DefaultTheme : Theme {
         }
 
         override val variants = object : InputFieldVariants {
+            private val basic: Style<BasicParams> = {
+                lineHeight { normal }
+                radius { normal }
+                fontWeight { normal }
+                paddings { horizontal { small } }
```
chausknecht, line 1353: Is already part of ``sizes`` property, thus it should be removed

styling/src/jsMain/kotlin/dev.fritz2.styling/theme/default.kt
```diff
@@ -1346,12 +1346,54 @@ open class DefaultTheme : Theme {
         }
 
         override val variants = object : InputFieldVariants {
+            private val basic: Style<BasicParams> = {
+                lineHeight { normal }
```
chausknecht, line 1350: At least ``lineHeight`` is better suited to the ``sizes`` property.

Proposal:
- add a ``basic`` styling for sizes proerty implementation too and call it from the three sizes

chausknecht (changes requested):

chausknecht (changes requested):
Please remember to merge current master; without KitchenSink won't compile.

chausknecht (approved):

PR #299, opened 2021-03-05
Make annotation processor for lenses aware of ctor present properties only

### LINKED ISSUE #295: Lenses annotation processor does not stick to ctor only properties, thus compilation fails for none ctor properties
Assume the folloowing data class as a Lense:
```kotlin
@Lenses
data class Cell(val id: Int, val symbol: String, val isInWinningGroup: Boolean = false) { // some props in ctor
    val isBlank: Boolean // property not in ctor!
        get() = symbol.isBlank()
}
```

The annotation processor generates the following piece of code (without comments of course):
```kotlin
object L {
  object Cell {
    val id: Lens<dev.fritz2.examples.tictactoe.Cell, Int> = buildLens("id", { it.id },{ p, v ->
        p.copy(id = v)})

    // This is the source of the failure! As the property is not part of the ctor, it can't be copied!
    val isBlank: Lens<dev.fritz2.examples.tictactoe.Cell, Boolean> = buildLens("isBlank", {
        it.isBlank },{ p, v -> p.copy(isBlank = v)})

    val isInWinningGroup: Lens<dev.fritz2.examples.tictactoe.Cell, Boolean> =
        buildLens("isInWinningGroup", { it.isInWinningGroup },{ p, v -> p.copy(isInWinningGroup =
        v)})

    val symbol: Lens<dev.fritz2.examples.tictactoe.Cell, String> = buildLens("symbol", { it.symbol
        },{ p, v -> p.copy(symbol = v)})
  }
}
```

This will lead to the following compilation error:
```
> Task :tictactoe:compileKotlinJs FAILED
8 actionable tasks: 3 executed, 5 up-to-date

Waiting for changes to input files of tasks... (ctrl-d then enter to exit)
e: C:\Users\chha\projects\external\oesb\fritz2-examples\tictactoe\build\generated\source\kaptKotlin\main\dev\fritz2\examples\tictactoe\GeneratedLenses.kt: (16, 39): Cannot find a parameter with this name: isBlank
```

This example is taken from the tictactoe example from the [fritz2-examples project](https://github.com/jamowei/fritz2-examples). You can find the data class within the [``model.kt``](https://github.com/jamowei/fritz2-examples/blob/master/tictactoe/src/commonMain/kotlin/dev/fritz2/examples/tictactoe/model.kt).

# Solution
The plugin should filter only properties present within the ctor of a class and only generate lenses for those!

The following part shows the possible source of the problem (``LensesAnnotationProcessor``):
```kotlin
  private fun handleDataClass(element: TypeElement, classData: TypeSpec): TypeSpec {
    val classSpec = TypeSpec.objectBuilder(classData.name ?: "unknown")
    // Here all properties of a class are handled; 
    // I think a filter could be added, which selects only those, that appears within the ctor!
    classData.propertySpecs.forEach { propertyData ->
      //FIXME: replace deprecated function call
      classSpec.addProperty((handleField(element.asClassName(), propertyData)))
    }
    
    return classSpec.build()
  }
```

chausknecht:
- Fix #295 
- First draft
- needs discussion; probably one could restructure the whole approach to sit on top of those kotlin metadata specific types instead of the more universal java types which lacks informations!


chausknecht:
We should at least add an exception if a data class is defined outsied of any package, as compilation fails with irritating error:
```
e: C:\Users\chha\projects\external\oesb\fritz2-template\build\generated\source\kaptKotlin\main\GeneratedLenses.kt: (10, 49): Type mismatch: inferred type is Lens<L.Cell, Int> but Int was expected
```
That is because now the generated ``object`` named as the data class leads to a name clash:
```kotlin
// without package for data class
object L {
  object Cell {
    val id: Lens<Cell, Int> = buildLens("id", { it.id },{ p, v -> p.copy(id = v)})
    //           ^^^^
    //           The compiler associates this type with the object itself which leads to the error
  }
}

// with package for data class
object L {
  object Cell {
    val id: Lens<dev.fritz2.template.Cell, Int> = buildLens("id", { it.id },{ p, v -> p.copy(id = v)})
    //           ^^^^^^^^^^^^^^^^^^^^^^^^
    //           the compiler can now distinguish the object from the target data class!
    //           so we need a package to the rescue!
  }
}
```

This is especially disturbing, as our [template project](https://github.com/jwstegemann/fritz2-template) does not offer a data class prepared for evaulating and experimentating with lenses, so a fritz2 newbie could easily get caught in this trap.

jwstegemann (approved):

PR #310, opened 2021-03-15
Decouple Modal and Toast components from local RenderContext

chausknecht:
Currently there are two components within fritz2's components library that are rendered independently from the current ``RenderContext`` of the main UI:
- [Modals](https://components.fritz2.dev/#Modal)
- [Toasts](https://components.fritz2.dev/#Toast)

Up until fritz2 version 0.9 the API has *forced* the client to declare those components from within a ``renderContext`` such as:
```kotlin
render() {
    clickButton { // here there is a (local) RenderContext
        text("Show")
    } handledBy toast {
        content { /* and so on */ }
    }
}
```

This clearly works, but it prohibits the declaration from none directly UI inetgrated code sections like ``Handlers``.
On top the implicitly injected ``RenderContext`` is not needed by the internal rendering of those components either, as those components *manage* their own context.

With fritz2 version 0.9.1 we improve the API by removing the ``RenderContext`` receiver from the factory functions of those components:
* ``showToast``
* ``toast``
* ``modal``

They can now be used completly outside of any local render context like this:
```kotlin
class ButtonStore : RootStore<Unit>(Unit) {
    val loading = tracker()

    val showMsg = handle {
        loading.track("running...") {
            delay(3000)
            modal { // called outside of any RenderContext!
                p { +"Process finished" }
            }
        }
    }
}

// and somewhere else:
render() { // here there is a (local) RenderContext
    clickButton {
        text("Show")
    } handledBy showMsg {  }
}
```

Of course they can still be called from within a local render context of course, so this is not an API breaking change and all application code should work as before.

#### Implementation notes

There is a new Interface for such *managed* component classes available: ``ManagedComponent<T>``. This should be implemented by all components which manage their own, *internal* RenderContext like toasts and modals until now.

Its ``render`` function simply omits the ``RenderContext`` parameter as known from ``Component<T>`` interface used for all other components.

Additionally it integrates the renamed static function ``managedRenderContext`` in order to create such an internal render context.


chausknecht:
Prepared a release notes ready comment

jamowei (approved):

Issue #323, opened 2021-03-19
Add Kotlin `@DslMarker` to components


jamowei:
Currently it's quite hard to get right auto-completion for relevant properties in the context, since all parent props are leaking in.
To fix that the components need to use the `@DslMarker` annotation.

haukesomm:
I just opened PR #328 to work on this issue.

PR #328, opened 2021-03-30
Add Kotlin @DslMarker to components

haukesomm:
This PR addresses issue #323.


jamowei:
We must use the `@HtmlTagMarker` from the fritz2 core to suppress the outer scope of `Tag`s in the component contexts.

PR #331, opened 2021-04-07
Using DSLMarkers for Components

jamowei:
On auto-completion for relevant properties in a `Component`-context no parent properties are leaking in anymore.


jamowei:
The `navBar` component was not using the `Component` interface. I fixed it in this PR also.

jamowei:
Tested with KitchenSink demo and works great!

jwstegemann (approved):
great. works as expected.

PR #338, opened 2021-04-13
Fixed problem with nested renders which contains conditions

### LINKED ISSUE #307: Problem with nested `render`s which contains conditions
Here is the code to reproduce:
```kotlin
fun main() {
    val switch = storeOf(false)
    val data = storeOf("Hello World!")

    render {
        switch(store = switch) {
            label("Switch")
        }
        hr {}
        switch.data.render {
            if (it) {
                data.data.render { text ->
                    if (text.isNotBlank()) p { +text }
                }
            }
        }
    }
}
```

jamowei:
Fixes #307


jamowei:
added failing test case

jwstegemann (approved):

PR #341, opened 2021-04-15
moves basicStyles of selectField and textArea to defaultTheme

mkempa-np:
moves basicStyles of selectField and textArea to defaultTheme, adds variant to textArea


chausknecht:
We need adations to the documentation in out [kitchensink project](https://github.com/jwstegemann/fritz2-kitchensink/pulls) for sure! This should be added in advance.

chausknecht (approved):
It seems a bit strange to me, that a "variant" context is introduced, even though there is only *one* available ;-)

But due to time restrictions and also because I see the similar use case to inputFields, this might be extended in the future, I just accept it without change.

Issue #292, opened 2021-03-04
Improve API of alert component


chausknecht:
**Note:** See [PR#361](https://github.com/jwstegemann/fritz2/pull/361) for breaking changes!

``alert`` offers an extension function on ``ComponentValidationMessage`` in order ease the process of emebdding messages into the UI. Currently this approach is used within our components by the ``formControl`` component, which has kinda dominated the API for its own needs, so it lacks important possibilities:

- the parameters should reflect *most* if not *all* *properties* of the ``AlertComponent``, besides the content and severity, which is defined by the message itself. All other aspects should be customizable!
- to avoid code duplication there should be some smart approach how to *reuse* the AlertComponent in a way.
- the current ``severity`` property of alert should be made private and hidden from the outside world.

haukesomm:
So I just stumbled over this issue and thought I might as well take care of it 😄 

> the parameters should reflect most if not all properties of the AlertComponent, besides the content and severity, which is defined by the message itself. All other aspects should be customizable!

> the current severity property of alert should be made private and hidden from the outside world.

Good point! I already thought something similar and I think we can get rid of the AlertComponent's `severity` property altogether by offering two new and more flexible properties instead: `color` (which should utilize the new `ColorScheme` class) and `icon`. 
This enables us to remove the irritating and unflexible `AlertSeverity` class from the theme and be more flexible with custom styling, both via the standard factory method as well as the `ComponentValidationMessage`-extension.

> to avoid code duplication there should be some smart approach how to reuse the AlertComponent in a way.

Is there anything in particular the could be done in your opinion?

haukesomm:
> the parameters should reflect most if not all properties of the AlertComponent, besides the content and severity, which is defined by the message itself. All other aspects should be customizable!

The `AlertComponent` needs some more changes in order to allow the above mentionend customization.

chausknecht:
The current improvement only deals with the central theming aspects.

The main focus of this issue is the offering of *configuration parameters* (like the styling) for the ``ComponentValidationMessage.asAlert``! This still lacks.

Have a look at the [TicTacToe]() example and the usage of the ``alert`` component for game messages like the winning or other ending messages. As there is a need to style the alert a little, it currently must be built by hand:
```kotlin
alert({
    // needed
    width { full }
    radius { small }
}) {
    // would be done by ``.asAlert()`` method, so it is cumbersome to craft this by hand
    severity { if (it.messages.first().severity == Severity.Success) success else info }
[...]

PR #356, opened 2021-04-26
Improve API of AlertComponent and use new color scheme

haukesomm:
This PR improves up the AlertComponent API by allowing more flexible customization via the `AlertSeverity`-style of the theme. Alerts mirroring the severity of a `ComponentValidationMessage` can still be created via the `ComponentValidationMessage.asAlert()` extension method. This PR also contains a lot of code cleanup in general.

Alerts now also utilize the new `ColorScheme` approach for color shades and typography.

---

**Breaking changes**

The actual usage of the `AlertComponent` does not change - the underlying styling within the theme does, however.
This means **if you are using a custom theme you need to update it** in order to implement the altered `AlertStyles` interface. In case you did not do any modifications to the `AlertComponent`'-theme your app should continue to work as intended.

- Updated `AlertSeverity` interface now includes support for the new `ColorScheme` class:
```kotlin
interface AlertSeverity {
    val colorScheme: ColorScheme
    val icon: IconDefinition
}
```
- `AlertVariants` now contains regular styles that are no longer split into styles for the background, text, icons, etc.:
```kotlin
interface AlertVariants {
    val subtle: BasicParams.(AlertSeverity) -> Unit
    val solid: BasicParams.(AlertSeverity) -> Unit
    val leftAccent: BasicParams.(AlertSeverity) -> Unit
    val topAccent: BasicParams.(AlertSeverity) -> Unit
    val discreet: BasicParams.(AlertSeverity) -> Unit
}
```
- Removed `AlertVariantStyles` interface

---

Solves #292 .


WestHuus (approved):
Nice job.

PR #380, opened 2021-05-03
new `linkButton` component

jamowei:
Added new `linkButton` to fritz2 components


WestHuus (commented):
On the one hand a nice work and a helpful component but on the other hand the question why we wrap the html <button> with the <a> when we can use the styling . Yes we'll have some more code but less in the dom.

WestHuus (approved):
After a discussion we decided to use this actual rendering of html elements.

PR #387, opened 2021-05-07
Add Slider Component

### LINKED ISSUE #386: Add Slider component
The should consider a Slider component, like [this one](https://chakra-ui.com/docs/form/slider).

Focus should be the ability to:
- define steps (by function and also maybe by discrete set)
- choose icon for sliding bubble

chausknecht:
* Add a slider component to fritz2.
* For details and documentation see [KitchenSink-App](https://components.fritz2.dev/#Slider)
* Basic usage:
    ```kotlin
    val valueStore = storeOf(100)
    slider(value = valueStore) {}
    ```

fix  #386

### DIFF
--- a/styling/src/jsMain/kotlin/dev/fritz2/styling/theme/default.kt
+++ b/styling/src/jsMain/kotlin/dev/fritz2/styling/theme/default.kt
@@ -50,9 +50,9 @@ open class DefaultTheme : Theme {
             override val gray800 = "#1A202C"
             override val gray900 = "#171923"
 
-            override val neutral = // white
+            override val neutral =
                 ColorScheme(
-                    main = "#ffffff",
+                    main = "#FFFFFF",
                     mainContrast = gray700,
                     highlight = gray100,
                     highlightContrast = gray700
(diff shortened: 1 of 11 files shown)


styling/src/jsMain/kotlin/dev/fritz2/styling/theme/default.kt
```diff
@@ -2068,4 +2068,116 @@ open class DefaultTheme : Theme {
...
             }
     }
+
+    override val slider = object : SliderStyles {
+
+        val self = this
+
+        override val sizes = object : FormSizes {
+            override val large: Style<BasicParams> = {
+                css("--sl-size: 1.5rem")
```
WestHuus, line 2078: I would insert a second css var for the spacing that is used. So you have for every size a configurable spacing.
Further in styling you can use the css-var instead of the value and  it will be forgotten when you wanna change the spacing.

jamowei (approved):

styling/src/jsMain/kotlin/dev/fritz2/styling/theme/default.kt
WestHuus: solved in https://github.com/jwstegemann/fritz2/pull/387/commits/c03de92707ffae917d5de095fda227938a177780

PR #407, opened 2021-05-27
Inserted calling of `styling` parameter of `alert` component

### LINKED ISSUE #406: Alert Component: Custom styling is not applied
```
                    alert({
                            background { color { "red" } }
                        }) {
                        content("The custom styling is not applied to the component and it appears in the theme's info color.")
                    }
```
Severity background colors work fine - this is about any styling.

WestHuus:
fixes #406

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
@@ -149,6 +149,8 @@ open class AlertComponent : Component<Unit> {
                     AlertVariant.TOP_ACCENT -> Theme().alert.variants.topAccent
                     AlertVariant.DISCREET -> Theme().alert.variants.discreet
                 }.invoke(this, this@AlertComponent.severity.value(Theme().alert.severities).colorScheme)
+
+                this as BoxParams; styling()
             }) {
                 box(styling = {
                     css("margin-right: var(--al-icon-margin)")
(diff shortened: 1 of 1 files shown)


jamowei (changes requested):
Do we need a `flexBox` here? When not it's would be nicer in my opinion to use a `div` instead and then no casting is needed to call the `styling()` function ;-)

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
```diff
@@ -163,7 +164,7 @@ open class AlertComponent : Component<Unit> {
                     }
                 }
 
-                box(baseClass = alertContentCss) {
+                div(baseClass = alertContentCss.name) {
```
chausknecht, line 167: Imho wrong! We need a ``StyleClass`` there and not a ``String`` - as the class also gets applied, I would definitly consider this as false correction!

chausknecht (changes requested):
I think we should cancel this PR in favour of PR #361, which already solves the problem.

components/src/jsMain/kotlin/dev/fritz2/components/alert.kt
chausknecht: Ok, I revoke my comment!

This is for sure something off within our API... the overloaded factories have a different type of course, as the styling DSL (and therefor the type ``StyleClass``) is not available at core level!

chausknecht:
Obsolete, PR #361 covers this already!

Issue #404, opened 2021-05-26
Investigation of some DataTable problems


chausknecht:
We should investigate the following unexpected and annoying behaviours:
- how to realize a smooth editable table without _flickering_ as the whole row gets deleted and newly rendered even if the ID does not change, but only the content of *one* cell!
- how to preserve the _tab order_ for an editable cell.
- try to find ``Flow`` that reduces the speed since the finalization of the table. Probably this is related with too aggressive reaction to changes. Some cache / surpressing of unchanged data could be helpful there.

metin-kale-cf:
#411 solves this issue

chausknecht:
Unfortunately PR #411 does not work! We have to consider that a possible activated sorting will change the index, which must be reflected by renderEach.

I have tried to combine both approaches as follows:

val indexedRowIdProvider: (IndexedValue<T>) -> Int = { (index, value) -> 
    rowIdProvider(value).hashCode() * 31 + index.hashCode() 
}

Unfortunately this also has improved the situation of the editable DataTable examples within the KitchenSink only partly!

In fact it is possible that we must delegate more responsability on top of the client for providing an efficient yet sufficient rowIdProvider, which reflects the need of the table:

    often the id of the model is sufficient
    if all / many columns are editable, their property counterparts need probably to be included also

We must investigate those concerns and find a solution and also probably some recipes for the users.

chausknecht:
The core of this problem and also the "flickering" when sorting the table is with high probability located in the outer ``renderEach`` call of the content rendering method: We absolutly must remove the index (and other changing properties) from the id providing. Only the ID should really be responsible for identifying the ``<tr>`` element, so it can be just _moved_!

The sorting order should be simply applied by the appearence of the rows within the flow given into ``renderEach`` (which is sorted before).

In order to provide the current _index_ to each cell, we must combine the index with the row content in the inner ``renderEach``-Block! That means there is the need to store the ``id -> index`` relation within a separate store or as extension of the datamodel. This way the index can be provided to each cell and be used for styling purposes for example.

Pitel:
Maybe the flickering has simple solution: [debounce](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/debounce.html) with short (50 ms?) timeout.

chausknecht:
> Maybe the flickering has simple solution: debounce with short (50 ms?) timeout.

[...]

PR #411, opened 2021-05-31
datatable: call renderEach using rowIdProvider instead of hashCode

metin-kale-cf:
Using hashCode as rowProvider for renderEach() results in useless rerenderings which cause some issues.

when using a dataTable with inputFields the table flickers after each change and focusing next element with the TAB-Key does not work either. Using the rowIdProvider solves both problems.

Example:
Data.kt

    @Lenses
    data class Data(val id: Int = 0, val firstname: String = "", val lastname: String = "")

Example.kt

            // ...
            dataTable(rows = store, rowIdProvider = Data::id) {
                columns {
                    // ...
                    column(title = "Firstname") {
                        lens(L.Data.firstname)
                        content { value, store, _ ->
                            inputField(value = store) {}
                        }
                    }
                    column(title = "Lastname") {
                        lens(L.Data.lastname)
                        content { value, store, _ ->
                            inputField(value = store) {}
                        }
                    }
                }
                // ...
            }
            // ..

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/datatable/component.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/datatable/component.kt
@@ -403,7 +403,7 @@ open class DataTableComponent<T, I>(val dataStore: RootStore<List<T>>, protected
                 styling()
             }) {
                 component.stateStore.renderingRowsData(component)
-                    .renderEach(IndexedValue<T>::hashCode) { (index, rowData) ->
+                    .renderEach({ rowIdProvider(it.value) }) { (index, rowData) ->
                         val rowStore = component.dataStore.sub(rowData, rowIdProvider)
                         val isSelected = this@DataTableComponent.selectionStore.isDataRowSelected(rowStore.current)
                         tr {


components/src/jsMain/kotlin/dev/fritz2/components/datatable/component.kt
```diff
@@ -403,7 +403,7 @@ open class DataTableComponent<T, I>(val dataStore: RootStore<List<T>>, protected
                 styling()
             }) {
                 component.stateStore.renderingRowsData(component)
-                    .renderEach(IndexedValue<T>::hashCode) { (index, rowData) ->
+                    .renderEach({ rowIdProvider(it.value) }) { (index, rowData) ->
```
chausknecht, line 406: Unfortunately this does not work! We have to consider that a possible activated sorting will change the index, which must be reflected by ``renderEach``. 

I have tried to combine both approaches as follows:
```kotlin
val indexedRowIdProvider: (IndexedValue<T>) -> Int = { (index, value) -> rowIdProvider(value).hashCode() * 31 + index.hashCode() }
```
Unfortunately this also has improved the situation of the editable DataTable examples within the KitchenSink only partly!

In fact it is possible that we must delegate more responsability on top of the client for providing an efficient yet sufficient ``rowIdProvider``, which reflects the need of the table:
- often the _id_ of the model is sufficient
- if all / many columns are editable, their property counterparts need probably to be included also

We must investigate those concerns and find a solution and also probably some **recipes** for the users.

chausknecht (changes requested):
We sould discuss this further in issue #404

PR #421, opened 2021-06-04
Repair centered modal

### LINKED ISSUE #399: Fix centering of `modal`
Possible styling solution:
```kotlin
override val centered: Style<BasicParams> = {
    position {
        fixed {
            top { "50%" }
            left { "50%" }
        }
    }
    css("transform: translateX(-50%) translateY(-50%);")
}
```

chausknecht:
Modal with variant ``centered`` now works as expected, as both axis are now included.

fixes #399


chausknecht:
The proposed solution does not work, as it dows not take the ``sizes`` property into account!

chausknecht:
Closed in favour of PR #442, which has the far better solution!

PR #422, opened 2021-06-07
Fixed self type loss after using global functions on `Listener` like `preventDefault` etc.

jamowei:
Fixes the following code example:
```kotlin
input {
    value(store.data)
    // not compiling, because of returning Listener instead of DomListener after using preventDefault()
    changes.preventDefault().values() handledBy store.update
}
```


WestHuus (approved):
Works fine. Simple solution :-)

PR #430, opened 2021-06-09
Make track safe for unsafe operations

### LINKED ISSUE #425: Unclear error handling
I'm learning Fritz2, and I'm making a login form. When user enters bad credentials, and the API returns HTTP 401, I'd like to show the error (toast). I can't find a way how.

Another related issue is in `Tracker`. When I start the remote request, and it fails, the button keeps spinning and I couldn't find any method how to stop it.

* Make special chapter in documentation about error handling.
* Demonstrate it in remotes example.
  * I have uBlock, which blocks the request to the domain you use in this example. I see the error in developer console, but nothing in the UI.

Abouty the tyracker problem, I'd suggest to change the `track` method like this:
```kotlin
// OLD (0.10.1)
return operation().also { state.value = null } 

// NEW
return try { operation() } finally { state.value = null }
```

I couldn't find any. Just provide some documentation and example.

**Additional context**
My handler looks like this. If you could show me how to display the error toast, I'd be grateful.

```kotlin
val login = handle { model ->
    loading.track {
        if (validator.isValid(model, Unit)) {
            try {
                http("/apiIs2/authUser")
                    .acceptJson()
                    .contentType("application/json")
                    .body(
                        JSON.stringify(
                            json(
                                "username" to model.login, "password" to model.password
                            )
                        )
                    )
                    .post()
                    .getBody()
                router.navTo("main")
            } catch (t: Throwable) {
                console.log(t)
                toast { // NOT DISPLAYED :(
                    content {
                        alert {
                            title("Login error")
                            content(
                                if (t is FetchException && t.statusCode == 401.toShort()) {
                                    "Bad credentials"
                                } else {
                                    "$t"
                                }
                            )
                        }
                    }
                }
            }
        }
        model
    }
}
```

chausknecht:
Currently the tracker does expect safe operations in order to work, so the client has to make shure, that no exception is beeing thrown by the passed operation.

This is quite dangerous, as the tracker will not stop working by itself (and there would be the need for a kind of _reset_ handler to be called when dealing with the exception).

This example should work now: The tracker will be stopped, but the exception is not swallowed and could be handled appropriately where needed:
```kotlin
val fooStore = object : RootStore<Int>(0) {
    val t = tracker()

    val h = handle {
        t.track {
            // unsafe operation is passed to ``track``
            delay(500)
            throw Exception("Boom!")
            it + 1 // will never be reached
        }
    }
}
```

fixes #425

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/tracking/tracking.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/tracking/tracking.kt
@@ -36,12 +36,19 @@ class Tracker(
     /**
      * Tracks a given [operation].
      *
+     * Works also with unsafe operations that throw exceptions, as the tracking gets stopped. The exceptions are
+     * not swallowed though.
+     *
      * @param transaction text describing the transaction
      * @param operation function to track
      */
     suspend fun <T> track(transaction: String = defaultTransaction, operation: suspend () -> T): T {
         state.value = transaction
-        return operation().also { state.value = null }
+        return try {
+            operation()
+        } finally {
+            state.value = null
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jwstegemann (changes requested):
Looks good but please add a test.

jwstegemann (approved):

PR #435, opened 2021-06-10
Added new `reset()` function to `Validator`

### LINKED ISSUE #432: Extend `Validator` with `reset()` function and make internal `state` protected instead of `private`
Currently there is no easy way to reset the list of validation-messages in the validator. Therefore we should provide a `reset()` function and open up the `state` for creating own validating functions by the implementer.
```kotlin
actual abstract class Validator<D, M : ValidationMessage, T> actual constructor() {

    private val state = MutableStateFlow<List<M>>(emptyList())
    ....
}
```

jamowei:
```kotlin
val save = handleAndEmit<Person> { person ->
    // only update the list when new person is valid
    if (validator.isValid(person, "add")) {
        emit(person)
        validator.reset()
        Person()
    } else person
}
```
Fixes  #432

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
@@ -31,6 +31,15 @@ actual abstract class Validator<D, M : ValidationMessage, T> actual constructor(
         get() = state.value
 
     /**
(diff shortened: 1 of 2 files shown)


WestHuus (approved):
Works fine.

Issue #448, opened 2021-06-16
Add tertiary and neutral color options to PushButtonTypes


metin-kale-cf:
https://github.com/jwstegemann/fritz2/blob/71d13d7888de1914c537c4c93ec071c81ea90fb1/styling/src/jsMain/kotlin/dev/fritz2/styling/theme/definitions.kt#L438

PushButtonTypes should contain tertiary and neutral colors

Currently i have to use

    type { Theme().colors.neutral }

to use neutral or tertiary colors.

It might be also considered to replace the `PushButtonTypes` Interface with the existing `Colors` Interface, because the return type is already limited to ColorScheme, so that the ColorProperties cannot be used anyway.

jwstegemann:
Indeed, thank you. Would you mind offering a PR? ;-)

chausknecht:
> It might be also considered to replace the PushButtonTypes Interface with the existing Colors Interface, because the return type is already limited to ColorScheme, so that the ColorProperties cannot be used anyway.

We had the clear intention to **not** expose some coloring there and focus on a more *semantic* meaning to those types.

I would prefer discussing this issue therefore! Also it was clearly intended to not support *all* colorschemes, but only those with strong semantic meaning.

It is possible that we failed to make our intension clear, but simply changing this without discussion could lead to a ping pong effect!

So better wait until our team is back from holidays?

PR #449, opened 2021-06-16
added neutral and tertiary colors to PushButtonStyles #448

metin-kale-cf:
see https://github.com/jwstegemann/fritz2/issues/448


chausknecht:
As we will rework our support for coloring components, I close this PR.

See Proposal #459 instead!

PR #451, opened 2021-06-18
fixed slider styling with delegations

metin-kale-cf:
I tried to override the Slider Themes in a custom Theme using:

    private val _slider = super.slider
    override val slider: SliderStyles = object : SliderStyles by _slider {
        override val horizontal: SliderCoreStyles = object : SliderCoreStyles by _slider.horizontal {
           ...
        }
    }

But it did not work. I realized, that the problem is with the SliderStyles.core(...) in:
https://github.com/jwstegemann/fritz2/blob/71d13d7888de1914c537c4c93ec071c81ea90fb1/styling/src/jsMain/kotlin/dev/fritz2/styling/theme/definitions.kt#L734

The Problem is that when using delegations, we actually have two SliderStyles instances, the base instance, and the custom instance which uses delegations to extend the base instance. 
The SliderStyles.core(...) runs in the base implementation instead of the custom implementation and returns the wrong instance of the SliderCoreStyles.

Here is a minimal example of the problem https://pl.kotl.in/00foZksY1

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/slider/component.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/slider/component.kt
@@ -17,10 +17,7 @@ import dev.fritz2.styling.name
 import dev.fritz2.styling.params.BasicParams
 import dev.fritz2.styling.params.BoxParams
 import dev.fritz2.styling.params.Style
-import dev.fritz2.styling.theme.FormSizes
-import dev.fritz2.styling.theme.IconDefinition
-import dev.fritz2.styling.theme.Icons
-import dev.fritz2.styling.theme.Theme
+import dev.fritz2.styling.theme.*
 import kotlinx.coroutines.flow.*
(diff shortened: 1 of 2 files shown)


chausknecht:
Nice catch!

But I do not like the call of ``Theme()`` within this ``definition.kt`` - it kinda smells because of cyclic dependency.

Wouldn't it be better to request a ``theme`` instance as parameter? So make a "static" method out of it via companion object? Imho that would be a cleaner solution.

metin-kale-cf:
> Wouldn't it be better to request a `theme` instance as parameter? So make a "static" method out of it via companion object? Imho that would be a cleaner solution.

What about a extension function for core()? The usage will be the same, but it will always use the correct SliderStyles without needing a extra parameter. 

I just changed it, if you still prefer passing Theme(), i could do that

chausknecht (approved):
Looks good!

PR #452, opened 2021-06-18
Add formGroup component

### LINKED ISSUE #440: Find solution for grouping formControls with horizontal layout (label + form)
We should introduce a grouping component for FormControls, which controls the overall layout structure. This would enable the following features:
- placing the label on the left side of the control (currently we only support the vertical placement)
- adapting the width of the left hand side label space to the situation needs for *all* controls within such a group.

```
formGroup {
    label { 
        placement { top } // or left -> top should remain default in theme
        width("200px") // or any sizing value; default is ``auto``, width only applied for left placement
    }

    // the placed formControls adapt to the enforced layout
    items {
        formControl {
        }
        formControl {
        }
        ...
    }
}
```

Consider placements based upon breakpoints:
```
formGroup {
    labels(
        // on small devices prefer vertical layout
        sm = {
            placement { top } 
        },
        // starting from medium sized devices put lables on the side
        md = {
            placement { left }
            width("20rem") 
        }
    )
}
```

### LINKED ISSUE #445: FromControl label should have "for" attribute
**Describe the bug**
FromControl label should have "for" attribute, so when user click on the label, the input gets automaticaly focused.

**To Reproduce**
1. Create simple `formControl` with `label` and `inputField`.
2. Click on the label.
3. Nothing happens.

**Expected behavior**
Cursor will appear inside the `inputField` (or, in case of chackbox, the checbox is toggled)

**Additional context**
Or, you can nest the `<input`>` inside `<label>` to get the same behaviour.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

chausknecht:
The new formGroup component is an additional grouping element for multiple formControls. It enables to configure the overall layout:
- place the labels on top (default and current behaviour of a  fromControl) or on the left side
- if placed left, enables to define the width of the label sections in an consistent way all over the whole form.
- could be extended in future versions to offer more capabilities (handle global form events and so on)

fixes #440 
fixes #445

### Small API breaking and Migration

The internal structure of the ``fromControl`` code has been changed. If an application creates its own derived ``FormControlComponent`` or custom ``ControlRenderer``, there is the need to import a new package:

```kotlin
import dev.fritz2.components.forms.control.*
```


chausknecht:
Consider to shorten IDs by:
- calling ``hashCode()`` on uniqueId
- applying shortening function ``generateAlphabeticName`` from styling module:
```kotlin
internal const val charsLength = 52

/* start at 75 for 'a' until 'z' (25) and then start at 65 for capitalised letters */
internal fun getAlphabeticChar(code: Int): Char = (code + if (code > 25) 39 else 97).toChar()

/* input a number, usually a hash and convert it to base-52 */
internal fun generateAlphabeticName(code: Int): String {
    val name = StringBuilder()
    var x = code
    while (x > charsLength) {
        x /= charsLength
        name.append(getAlphabeticChar(x % charsLength))
    }
    return name.toString()
}
```
-> move this into core

jamowei (approved):

PR #483, opened 2021-07-12
Deprecate `box` in favor of `div`

### LINKED ISSUE #379: Consider to deprecate box component
As standard HTML tags can now (Issue #337) be styled using the styling DSL, we could consider to deprecate the ``box`` component, which is just a styled div!
```kotlin
// before 0.10
box({
    // some styling DSL
}) {...}

// starting with 0.10. absolutly equivalent
div({
    // some styling DSL
}) {...}
```

haukesomm:
This PR deprecates the `box` factory method in favor of `div` because, in combination with fritz2.styling, it offers the exact same functionality. All calls of `box` will have to be replaced eventually.

Closes #379 .

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/box.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/box.kt
@@ -33,13 +33,7 @@ import dev.fritz2.styling.params.GridParams
  * @param prefix the prefix for the generated CSS class resulting in the form ``$prefix-$hash``
  * @param content a lambda expression for setting up the content and events of the ``div`` element itself
  */
-@Deprecated(
-    "Use 'div' instead of 'box' (same functionality)",
-    ReplaceWith(
-        expression = "div(styling, baseClass, id, prefix) { content() }",
-        imports = ["dev.fritz2.styling.div"]
-    )
-)
+@Deprecated("Use 'div' instead of 'box' (same functionality)")
 fun RenderContext.box(
     styling: FlexParams.() -> Unit = {},
     baseClass: StyleClass = StyleClass.None,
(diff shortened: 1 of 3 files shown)


components/src/jsMain/kotlin/dev/fritz2/components/box.kt
```diff
@@ -33,6 +33,13 @@ import dev.fritz2.styling.params.GridParams
  * @param prefix the prefix for the generated CSS class resulting in the form ``$prefix-$hash``
  * @param content a lambda expression for setting up the content and events of the ``div`` element itself
  */
+@Deprecated(
```
chausknecht, line 36: I think the replacement expression is far too complicated and wrong:
```kotlin
box({
    background { color { "purple" } }
}) {
    +"Test"
}
```
If applied the above becomes to:
```kotlin
div(fun FlexParams.() {
    background { color { "purple" } }
}, StyleClass.None, null, "box") {
    +"Test"
}
```
Definitly not correct!

Should it not simplex be a chnage from `box` to `div` like this?
```kotlin
div({
    background { color { "purple" } }
}) {
    +"Test"
}
```
The import is nice of course!

chausknecht (changes requested):
Correct replacement expression if my assumptions is correct!

chausknecht:
I think a corresponding PR for the KitchenSink is yet missing - lots of boxes left there ;-)

components/src/jsMain/kotlin/dev/fritz2/components/box.kt
haukesomm: Afaik `ReplaceWith` does not have full support for both optional parameters and lambdas which results in the generated replacement being overly complicated, yet technically correct. The only other option that comes to my mind is to get rid of the `ReplaceWith` as the conversion is simple enough to just be mentioned in the release notes 🤔

chausknecht (approved):

PR #489, opened 2021-07-16
Some Improvements and Bugfixing around Forms

### LINKED ISSUE #429: FormControl.textArea still uses 'store' as parameter name
In
https://github.com/jwstegemann/fritz2/blob/f760ad579daf8ddff8535eb1f7e73b28b898643d/components/src/jsMain/kotlin/dev/fritz2/components/formcontrol.kt#L268

`store` parameter name should be `value`

in a RenderContext the textArea can be used like this:

    textArea(value = store){}

But in the FormControl the textArea parameter was not changed to `value = ` with the 0.10 Release, it still uses `store = ` as parameter name

    formControl {
        label("Label")
        textArea(store = store){}
    }

### LINKED ISSUE #457: Correct small internal coding issues within FormControlComponent
- leaking ``this`` in constructor (``init`` block; registration of renderers)
- change sizes from small to uppercase

### LINKED ISSUE #472: Consider to enhance the label of a formControl to be a Flow<String> rather than simply a String
## Motivation:
If a label depends on the state of some other form component's state, currently the whole formControl must be rerendered.

## Proposal
It would be quite easy (and only API enhancing only) to replace the current label's type from
```kotlin
// currently
val label = ComponentProperty("")

// solution
val label = DynamicComponentProperty<String>(emptyFlow())
```

## Current vs Future usage
The usage in the default renderes is as follows:
```kotlin
// SingleControlRenderer
label({
    component.labelStyle.value()
}) {
    component.controlRegistration.assignee?.id?.let { `for`(it) }
    className(formGroupElementLabelMarker)
    component.label.values.asText() // would change!
}

// ControlGroupRenderer
legend({
    component.labelStyle.value()
}) {
    className(formGroupElementLegendMarker)
    component.label.values.asText() // would change!
}
```

### LINKED ISSUE #473: Repair Id assignment (for label in FormControls) for SelectField
Currently the SelectField component is faulty concerning the Id assignment. It does not assign the Id to the `select` tag itself, so a corresponding label won't be tied to this control.

### Check also all the other forms too
- issue #445 should have closed this already; obviously some controls still do not handle theis passed Ids not correctly!

### LINKED ISSUE #484: Change Sizes facility of form components to respect Item 502 of EfC
Currently the sizes of form controls are defined like this:
```kotlin
interface FormSizes {
    val small: Style<BasicParams>
    val normal: Style<BasicParams>
    val large: Style<BasicParams>
}


/**
 * definition of the theme's checkbox
 */
interface CheckboxStyles : SeverityAware {
    val sizes: FormSizes
    val input: Style<BasicParams>
    val icon: Style<BasicParams>
    val label: Style<BasicParams>
    val default: Style<BasicParams>
    val checked: Style<BasicParams>
}
```
With respect to [Item 502](Effective-fritz2-Components#item-502-strive-for-consistent-api-within-components-styling-facilities) we should change this.

chausknecht:
### Improve FormControl code

- make `FormSizeSpecifier` values uppercase, so more Kotlin alike
- improve render strategies initialization and customization: Instead of setting all up within the `init` block, there is now a new private method `initRenderStrategies` that produces the mapping and is directly called at beginning of the rendering process. This way no `this` pointer is leaked as before. For custom implementations there is now a protected hook function `finalizeRenderStrategies` that can be used to extend or change the renderer strategies!

#### Migration Guide:
If a custom renderer should be applied or a new factory should be added, custom implementation of `FormControlComponent` must replace the old `registerRenderStrategy` within the `init` block by a new mechanism:
```kotlin
// old way, no more possible!
class MyFormControlComponent : FormControlComponent {
    init {
        registerRenderStrategy("radioGroupWithInput", ControlGroupRenderer(this))
    }
}

// new way, override `finalizeRenderStrategies` method:
class MyFormControlComponent : FormControlComponent {

    // some new factory: `creditCardInput` with key of same name; used with `SingleControlRenderer`

    // another new factory: `colorInput` with key of same name and new renderer

    override fun finalizeRenderStrategies(
        strategies: MutableMap<String, ControlRenderer>,
        single: ControlRenderer,
        group: ControlRenderer
     ) {
         // override setup for a built-in factory:
         strategies.put(ControlNames.textArea, MySpecialRendererForTextAreas(this))

         // register new factory
         strategies.put("creditCardInput", single)

         // register new factory with new renderer
         strategies.put("colorInput", ColorInputRenderer(this))
     }
}
```

fixes #457

### Improve sizes aspect for components in theme

- in respect to [EfC 502](Effective-fritz2-Components#item-502-strive-for-consistent-api-within-components-styling-facilities) a new mixin alike interface `FormSizesAware` has been introduced
- the `FormSizes` interface is renamed to `FormSizesStyles`

#### Migration Guide:
If a component supports a size property and relies on the old `FormSizes` interface, just rename the receiver type appropriate to `FormSizesStyles``:
```kotlin
// old
val size = ComponentProperty<FormSizes.() -> Style<BasicParams>> { Theme().someComponent.sizes.normal }

// new
val size = ComponentProperty<FormSizesStyles.() -> Style<BasicParams>> { Theme().someComponent.sizes.normal }
//                           ^^^^^^^^^^^^^^^
//                           choose new name
```

fixes #484

### Make FormControl Labels dynamic

- the `label` property of a FormControl now accepts also `Flow<String>`. Thus the label can dynamically react to some other state changing.

#### Migration Guide:
The code for a `ControlRenderer` implementation needs to be adapted. Use the following recipe:
```kotlin
class ControlGroupRenderer(private val component: FormControlComponent) : ControlRenderer {
    override fun render(/*...*/) {
        // somewhere the label gets rendered (label / legend or alike)
        label {
            // old:
            +component.label.value
            // change to:
            component.label.values.asText()
        }
    }
}
```

fixes #472

### Adapt FormControl's textArea function parameters

- the optional store parameter now also is named `value` like as within the main `textArea` factory.

fixes #429

### Repairs label behaviour for SelectField

- a click onto a label within a FormControl now directs to the SelectField.
- the id of a SelectField is now correctly set, so that a FormControl can now fill the `for` attribute of the label correctly.

fixes #473


chausknecht:
Requires PR #488

haukesomm (approved):
Looks very good to me and seems to work as expected 👍🏻

PR #494, opened 2021-07-22
Allow `sub()` on all `Store´s

metin-kale-cf:
Removes the `RootStore` dependency in the `SubStore`, so that `SubStore`s can be created by calling `sub()` function implemented in `Store` interface. Therefore it is now possible to create a `SubStore` from every `Store` object.

Because one `SubStore` generic type gets obsolete, it will break existing code, but it is easy to fix that. Just remove the first not type parameter.
```kotlin
// before
class SubStore<R, P, T> {}
// now
class SubStore<P, T> {}
```

```kotlin
// before
val addressSub: SubStore<Person, Person, Address> = store.sub(addressLens)
// now
val addressSub: SubStore<Person, Address> = store.sub(addressLens)
```


jamowei:
All tests are running, but there is an issue in validation.kt which needed a workaround.

Previously the functions validationMessage() and validationMessages() directly accessed the root store of the substore to access the validations. The SubStore does not hold the RootStore anymore so i had to add a findRootStore() Method as workaround which iterates through the parent stores to get the root. It should be discussed, whether there is a better way.

jamowei (approved):

PR #496, opened 2021-07-26
Add `PaperComponent` and `CardComponent`

### LINKED ISSUE #438: Move areas from Popover into an own component and use `DropdownComponent` to create popovers
In order to simplify things we should consider the following points about popovers and dropdowns:

- We should consider moving the below area structure from the `PopoverComponent` into an own component.
The new component could be embeddable within modals too.
- The new `DropdownComponent` could be used to render popovers in the future. 
- As it becomes a more versatile component, it might also be a good idea to find a better name for the `DropdownComponent` as described in #447 .

grafik

haukesomm:
This PR adds a new `CardComponent` that behaves similar to the __old__ `PopoverComponent`s content.

## Paper Component

Component displaying content in a card-like box that can either appear elevated or outlined and scales with the specified `size` of the component.

### Example

```kotlin
paper {
    size { /* small | normal | large */ }
    type { /* normal | outline | ghost */ }

    content {
        // ...
    }
}
```

Bildschirmfoto 2021-08-17 um 19 09 11

## CardComponent

A component displaying the typical sections of a card inside a `PaperComponent`.  
The available sections are a header, a footer and the actual content.

### Example

```kotlin
card {
    size { /* small | normal | large */ }
    type { /* normal | outline | ghost */ }

    header {
        // ...
    }
    content {
        // ...
    }
    footer {
        // ...
    }
}
```

Bildschirmfoto 2021-08-17 um 19 17 18

---

Closes #438 .


components/src/jsMain/kotlin/dev/fritz2/components/paper.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import dev.fritz2.components.paper.PaperComponent
+import dev.fritz2.dom.html.RenderContext
+import dev.fritz2.styling.StyleClass
+import dev.fritz2.styling.params.BasicParams
+
+/**
+ * This factory function creates a [PaperComponent].
+ *
+ * @param styling a lambda expression for declaring the styling of the toast using fritz2's styling DSL
```
chausknecht, line 11: Replace all "toast" occurences with "paper"

components/src/jsMain/kotlin/dev/fritz2/components/paper.kt
```diff
@@ -0,0 +1,23 @@
...
+ * @param baseClass optional CSS class that should be applied to the toast element
+ * @param id the ID of the toast element
+ * @param prefix the prefix for the generated CSS class of the toast element resulting in the form ``$prefix-$hash``
```
chausknecht, line 14: The pattern is not correct for this component

components/src/jsMain/kotlin/dev/fritz2/components/paper.kt
```diff
@@ -0,0 +1,23 @@
...
+ * @param build a lambda expression for setting up the component itself
+ */
+fun RenderContext.paper(
+    styling: BasicParams.() -> Unit = {},
+    baseClass: StyleClass = StyleClass.None,
+    id: String? = null,
+    prefix: String = "card",
```
chausknecht, line 21: It's a paper, not a card ;-)

styling/src/jsMain/kotlin/dev/fritz2/styling/theme/default.kt
```diff
@@ -1054,6 +1054,118 @@ open class DefaultTheme : Theme {
...
+            override val outline: Style<BoxParams> = {
+                border {
+                    color { neutral.highlight }
+                    style { solid }
+                    width { "var(--paper-line-width)" }
+                }
+            }
+
+            override val ghost: Style<BoxParams> = {
+                // TODO: Still needed?
```
chausknecht, line 1092: Yes, think of card within modal! So the TODO could be removed imho

chausknecht (changes requested):

chausknecht (approved):

PR #503, opened 2021-08-16
Fix problem with missing ids on components with `for` attribute (radio, checkbox, switch)

### LINKED ISSUE #501: Weird bug in FormControl with multiple Switch/Checkboxes
i created a blank fritz2 project using the template and added the components dependency to reproduce the issue with a minimal example:

the main()-Method looks like:

    fun main() {
        render {
            formControl {
                switch(value = storeOf(false)) {
                    label("Label")
                }
            }
    
            formControl {
                switch(value = storeOf(false)) {
                    label("Label")
                }
            }
        }
    }

when i click the second switch, the first switch will be toggled. i cant toggle the second switch at all. The same issue also occurs using checkboxes.

jamowei:
Fixes #501


metin-kale-cf (changes requested):
The code example from #501 works, but this commit breaks substore functionality. 

    formControl {
        switch(value = Store.sub(L.Model.abc)) {
            label("ABC")
        }
    }

the above formcontrol is not editable anymore. I have also checked the master branch, there it is working as expected.

metin-kale-cf:
the problem here is the label id, which becomes the same as the inputId, if the store has a non-null id

image

metin-kale-cf:
1. validation is broken because following line still uses an empty string instead of `null`
https://github.com/jwstegemann/fritz2/blob/3a5550567e226611da9a0cbd8a74ac816ab01c5c/core/src/commonMain/kotlin/dev/fritz2/identification/inspector.kt#L11

2. it might be a rare use case, but if i have two stores using the same data model and use their substores, both will have the same id (e.g. null.abc), so the same issue will occur again.

if this problem should also be addressed, the solution might be more tricky. Either the store ids have to be enforced, or generated directly in the store as default. But that would break validation. because the RootValidator needs to have the same id.

jamowei:
I spend some time to find practical solution for the `formControl` component to solve this issue. Now the neither the input components nor the `formControl` component needs to render ids (generated or from stores) to the DOM! Instead I registered a simple click-handler for the outer label of the `formControl` for focusing the inner control element. Therefore there is no duplicate id issue anymore. If you want to set an id you must do this explicitly.

metin-kale-cf:
I tried different use cases, the last change seems to fix all issues. Without rendering the id to the DOM different stores without id, substores and validation are working fine now.

jwstegemann:
Concept:

- every store has a unique id:
    - either one is provided by the user
    - or one is generated (Id.next() -> increment global long value)
- a component uses this id, if a store is provided by the user and no other id is explicitly given by the user
- if an id is necessary to work for the component and no store is provided (by using a plain flow) an id is generated by the component using the same funtion
- an inspector does not need to create an id

components/src/jsMain/kotlin/dev/fritz2/components/inputField/component.kt
```diff
@@ -93,18 +93,18 @@ open class InputFieldComponent(protected val valueStore: Store<String>?) :
...
         context: RenderContext,
         styling: BoxParams.() -> Unit,
         baseClass: StyleClass,
         id: String?,
         prefix: String
-    ) {
-        context.apply {
+    ): Input {
+        return with(context) {
             input({
```
chausknecht, line 108: Same issue with `id` as at checkbox!

components/src/jsMain/kotlin/dev/fritz2/components/radios/componentGroup.kt
```diff
@@ -124,16 +123,16 @@ open class RadioGroupComponent<T>(protected val items: List<T>, protected val va
...
 
-        context.apply {
+        return with(context) {
             div({
-                layoutOf(this@RadioGroupComponent.orientation.value(OrientationContext))()
                 verticalAlign { top }
                 css("-webkit-box-align: center;")
                 alignItems { center }
+                layoutOf(this@RadioGroupComponent.orientation.value(OrientationContext))()
             }, styling, baseClass, id, prefix) {
```
chausknecht, line 136: Same issue with `id` as at checkbox!

components/src/jsMain/kotlin/dev/fritz2/components/radios/componentSingle.kt
```diff
@@ -126,35 +125,26 @@ open class RadioComponent(protected val value: Store<Boolean>? = null) :
...
+                verticalAlign { top }
+                css("-webkit-box-align: center;")
                 this@RadioComponent.size.value.invoke(Theme().radio.sizes)()
                 // to "capture" the invisible, absolute positioned `input`, see `radioInputStaticCss`
                 position { relative {  } }
-            }, baseClass = baseClass, id = id, prefix = prefix) {
-                inputId?.let {
-                    `for`(inputId)
-                }
+            }, baseClass, prefix = prefix) {
```
chausknecht, line 137: Same issue with `id` as at checkbox!

components/src/jsMain/kotlin/dev/fritz2/components/selectField/component.kt
```diff
@@ -131,22 +131,20 @@ open class SelectFieldComponent<T>(protected val items: List<T>, protected val v
...
                     this@SelectFieldComponent.items.indexOf(selectedItem).let { if (it == -1) null else it }
                 } handledBy internalStore.update
 
-            div({}, styling, baseClass + staticCss, prefix = prefix) {
+            label(styling, baseClass + staticCss, prefix = prefix) {
                 select({
                     this@SelectFieldComponent.variant.value.invoke(Theme().select.variants)()
                     this@SelectFieldComponent.size.value.invoke(Theme().select.sizes)()
-                }, styling, baseClass + staticCssSelect, grpId) {
+                }, styling, baseClass + staticCssSelect, id) {
```
chausknecht, line 147: Same issue with `id` as at checkbox!

components/src/jsMain/kotlin/dev/fritz2/components/slider/component.kt
```diff
@@ -347,6 +345,7 @@ open class SliderComponent(protected val store: Store<Int>? = null) :
                 styling()
             }, baseClass, id, prefix) {
```
chausknecht, line 346: Same issue with `id` as at checkbox!

components/src/jsMain/kotlin/dev/fritz2/components/switch/component.kt
```diff
@@ -116,21 +115,20 @@ open class SwitchComponent(protected val value: Store<Boolean>? = null) :
...
         return with(context) {
             label({
+                display { inlineFlex }
+                alignItems { center }
                 this@SwitchComponent.size.value.invoke(Theme().switch.sizes)()
                 // to "capture" the invisible, absolute positioned `input`, see `switchInputStaticCss`
                 position { relative {  } }
-            }, baseClass = baseClass, id = id, prefix = prefix) {
-                `for`(inputId)
+            }, baseClass, prefix = prefix) {
```
chausknecht, line 125: Same issue with `id` as at checkbox!

components/src/jsMain/kotlin/dev/fritz2/components/switch/component.kt
```diff
@@ -116,21 +115,20 @@ open class SwitchComponent(protected val value: Store<Boolean>? = null) :
...
                 input({
                     Theme().switch.input()
                     children("&[checked] + div") {
                         this@SwitchComponent.checkedStyle.value()
                     }
-                }, baseClass = switchInputStaticCss, prefix = prefix, id = inputId) {
+                }, switchInputStaticCss, id, prefix) {
```
chausknecht, line 131: Not sure if this is the right place for the `id`?

components/src/jsMain/kotlin/dev/fritz2/components/textarea/component.kt
```diff
@@ -102,8 +102,8 @@ open class TextAreaComponent(protected val valueStore: Store<String>? = null) :
         baseClass: StyleClass,
         id: String?,
         prefix: String
-    ) {
-        context.apply {
+    ): TextArea {
+        return with(context) {
```
chausknecht, line 106: Same issue with `id` as at checkbox!

components/src/jsMain/kotlin/dev/fritz2/components/typeAhead/component.kt
```diff
@@ -240,16 +242,16 @@ open class TypeAheadComponent(protected val valueStore: Store<String>?, protecte
...
-        val proposalsId = "proposals-{${uniqueId()}}"
+        val proposalsId = "proposals-{${Id.next()}}"
 
-        context.apply {
+        return with(context) {
             (this@TypeAheadComponent.valueStore?.data
                 ?: this@TypeAheadComponent.value.values) handledBy internalStore.preselect
 
-            inputField(
+            val input = inputField(
```
chausknecht, line 254: Same issue with `id` as at checkbox!

core/src/commonMain/kotlin/dev/fritz2/identification/id.kt
```diff
@@ -1,8 +1,20 @@
 package dev.fritz2.identification
 
+object Id {
+    private const val defaultLength = 6
+    private val chars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz".toCharArray()
+
+    fun next(length: Int = defaultLength) = buildString {
+        for (i in 0 until length) {
```
chausknecht, line 8: How about `repeat(length) { append(chars.random()) }` ? Not sure whether your version might be faster though!

core/src/commonMain/kotlin/dev/fritz2/identification/inspector.kt
```diff
@@ -61,7 +62,7 @@ class SubInspector<R, P, T>(
     /**
      * generates the corresponding id
```
chausknecht, line 63: `path` not `id`

core/src/commonTest/kotlin/dev/fritz2/identification/inspector.kt
```diff
@@ -22,20 +22,20 @@ class InspectorTests {
     fun testInspectorIds() {
```
chausknecht, line 22: rename to `fun testInspectorPaths`?

core/src/commonTest/kotlin/dev/fritz2/identification/inspector.kt
```diff
@@ -22,20 +22,20 @@ class InspectorTests {
...
         val addressInspector = rootInspector.sub(addressLens)
-        assertEquals("start.address", addressInspector.id, "sub model id not correct")
+        assertEquals(".address", addressInspector.path, "sub model id not correct")
         assertEquals(rootData.address, addressInspector.data, "sub model data not correct")
 
         val streetInspector = addressInspector.sub(streetLens)
-        assertEquals("start.address.street", streetInspector.id, "sub sub model id not correct")
+        assertEquals(".address.street", streetInspector.path, "sub sub model id not correct")
         assertEquals(rootData.address.street, streetInspector.data, "sub sub model data not correct")
     }
```
chausknecht, line 41: rename to `testPathsByElementLens`

core/src/commonTest/kotlin/dev/fritz2/identification/inspector.kt
```diff
@@ -98,18 +98,18 @@ class InspectorTests {
             Person("p3", Address("p3s1"))
```
chausknecht, line 98: rename to `testPathsByPositionLens`?

chausknecht (changes requested):
Looks great overall! Mission accomplished from my pov!

KS compiles without any related warning, but needs some further refinements. I will review it seriously tomorrow!

core/src/commonMain/kotlin/dev/fritz2/identification/id.kt
jamowei: I would prefer the for loop ;)

chausknecht (approved):

PR #510, opened 2021-08-23
Optimize DataTable's reactive rendering

### LINKED ISSUE #404: Investigation of some DataTable problems

### LINKED ISSUE #509: Eliminate vertical flickering of default hover effect in DataTable
Improve hover effect, so that the hovering border does not change the height of a row anymore. This has caused to enlarge the rows beneath the hovered one which has lead to a vertical flickering effect which is annoying to the eyes and disturbing overall.

chausknecht:
The DataTable now reacts much smoother to changes of its content, so that:
- a row itself (the `tr` tag) is not deleted until the `id` of its row disappears
- a cell (the `td` tag) is only rendered if its `String` represenatation via its `Lens` changes or some of the state related properties like the selection state or the sorting

This improves the handling of editable cells, as one can now *tab* through a row without rerendering it and therefore keeping the focus stable so it advances by default to the next column.

Be aware that you must adapt to those changes, if your editable DataTable relies on changes to one arbitrary (maybe not rendered property!) field of a model. This will not work any longer and must be solved otherwise.

fixes #404
fixes #509

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/datatable/component.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/datatable/component.kt
@@ -410,7 +410,7 @@ open class DataTableComponent<T, I>(val dataStore: RootStore<List<T>>, protected
             }) {
                 this@DataTableComponent.stateStore
                     .renderingRowsData(this@DataTableComponent, rowIdProvider).let { renderingData ->
-                        renderingData.map { it.first }.renderEach(indexedRowIdProvider) { (index, rowData) ->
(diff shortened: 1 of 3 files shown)


haukesomm (approved):
Looks very well to me! 👍🏻

PR #522, opened 2021-09-07
 Expose open/close handlers or an external store for submenu component

### LINKED ISSUE #520: menu component: option to collapse submenu when another submenu is opened (accordeon)
We're using the new menu component with the appFrame and have a lot of links in the navigation area. Showing more than one expanded submenu takes up too much space.
 
We would like the option to collapse an expanded submenu when another submenu is expanded, allowing only one single submenu to show its subitems at a time. Currently we cannot customize the menu component with this function ourselves because the expand/collapse handlers of submenu are not available. 

Additional information: We want to navigate to an overview page AND expand the entries after the user clicks on the collapsed submenu. 

[Mockup](https://ibb.co/pLnHW4H)

jamowei:
Fixes #520

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/menu/component.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/menu/component.kt
@@ -1,5 +1,6 @@
 package dev.fritz2.components.menu
 
+import dev.fritz2.binding.Store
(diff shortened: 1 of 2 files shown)


EvaBS (approved):
Thank you :-)

PR #523, opened 2021-09-08
Fixed not working `disabled` property in `formControl` component

jamowei:


chausknecht:
Let's think about a different approach: Why not just remove the `FormProperties` from the `FormControlComponent`? This way there is no need to pass the state down and there is less chance for ambiguous declarations. As most other specific properties, you just control this within the controls contexts.

I have checked also the `ControlRenderer` implementations:  They do not rely on the `FormProperties` properties. So a removal would also not break anything for this aspect.

chausknecht (approved):

PR #526, opened 2021-09-23
fix registering mount points

jwstegemann:
fixes two problems:

1.  child elements were already appended by mount-points to the parent element during the mapping. The mount-method late only moved them to their right position when calling replace element. not the child - elements are not connect to their parent until the mount-function
2.  a mount point has to forward its parent to its children as context if this parent is another mount point. Only if it is a regular tag it has to create a new "virtual" RenderContext


jwstegemann:
This PR should not be merged. An alternative approach that fixes these and some more issues is provided by #530

Issue #405, opened 2021-05-27
nested render on tracker() throws UnsupportedOperationException in Browser


WestHuus:
See example Code
```kotlin
  @ExperimentalCoroutinesApi
fun main() {
    val store = object : RootStore<String>("...") {
        val running = tracker()
        val start = handle {
            running.track("init") {
                delay(1500)
                "store"
            }
        }

        init {
            start()
        }
    }
    val otherStore = object : RootStore<String>("") {
        val running = tracker()
        val start = handle {
            running.track("init") {
                delay(600)
                "otherStore"
            }
        }

        init {
            start()
        }
    }

    store.watch()
    otherStore.watch()

    //not works
    render {
        store.running("init").render {
            div {
                otherStore.running("init").render {
                    store.data.render {
                        span { +it }
                    }
                }
            }
        }
    }
 
    //works
/**
    render {
        store.running("init").render {
            div {
                otherStore.running("init").render {
                    div {
                        store.data.render {
                            span { +it }
                        }
                    }
                }
            }
        }
    }
*/
}
```

Pitel:
I have something similar, but without tracker:

```kotlin
fun RenderContext.foo() {
  val store = storeOf(false)

  div { // It's broken without this div!
    store.data.render {
      p { +"$it }
    }
  }

  clickButton {
    text("Toggle")
  }.map { !store.current } handledBy store.update
}
```

At first it looks ok. But after clicking the button, I get the `UnsupportedOperationException` and the text "true" appears after the button (while the inital "false" stays above the button).

(The code is heravily simplified and untested, )

jillesvangurp:
@Pitel thanks for the workaround, running into exactly this as well with fritz2. Adding a div around any `store.data.render` works

PR #530, opened 2021-10-01
Alternative approach to rendering

jwstegemann:
- simplifies a lot of rendering-related code
- solves all(?) issues related to placeholders, rendering-orders, etc.


chausknecht:
I just added one more issue (#405 ) that should be fixed by this.

PR #541, opened 2021-10-29
Upgrade to gradle 7.2

chausknecht:
- upgrade gradle from 6.8 to 7.2
- upgrade to Kotlin 1.5.31
- upgrade lots of dependencies to fitting / newest versions
- pay ATTENTION: kotlinpoet does not work from 1.10.x upwards (until 1.10.2 tested) -> only upgrade to 1.9.0
- remove `jcenter`  as it is deprecated
- make it possible to compile fritz2 with Java 16 and 17 theoretically; as KAPT relies on deprecated internal API of JDK (by version 9), it now fails to work with Java >= 16. So client applications can't create Lenses. See issue #542 for alternative proposal.


jamowei (approved):
Working on Linux!

PR #544, opened 2021-11-07
Rework RenderContext

jwstegemann:
RenderContext now is an Interface offering all functions to render things (dynamic and static)
A Tag is an implementation of this interface (adding everything to handle a DOM-element)
TagContext has been removed
The nearest MountPoint now is available on scope when rendering (some convenience methods to access it)
Allow the registration of lifecycle-handlers after mounting and before unmounting a Tag from the DOM:
```kotlin
div {
    afterMount(someOptionalPayload) { tag, payload ->
        //do something here
    }

    beforeUnmount(someOptionalPayload) { tag, payload ->
        //do something here
    }   
}
```


chausknecht (dismissed):

core/src/jsMain/kotlin/dev/fritz2/dom/mount.kt
```diff
@@ -221,26 +155,42 @@ fun <V> Flow<List<V>>.eachIndex(): Flow<Patch<V>> =
...
-                is Patch.InsertMany -> target.domNode.insertMany(patch.elements, patch.index)
-                is Patch.Delete -> target.domNode.delete(patch.start, patch.count) { node ->
-                    val job = jobs.remove(node)
-                    if (job != null) job.cancelChildren()
-                    else console.error("could not cancel renderEach-jobs!")
+                is Patch.Insert -> {
+                    target.domNode.insert(patch.element, patch.index)
+                    val mountPointImpl = mountPoints[patch.element.domNode]
+                    if (mountPointImpl != null) mountPointImpl.runAfterMounts()
+                    else console.error("could not run afterMount on inserting ${patch.element}")
```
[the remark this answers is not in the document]
jwstegemann: What would be missing is the mount-point. Since all renders (even the global one) are MPs and put themselves on the scope, this would mean that something regarding the scopes does not work. I don't think, that this could be fixed in user code.

chausknecht (dismissed):

core/src/jsMain/kotlin/dev/fritz2/dom/html/elements.kt
```diff
@@ -1315,19 +1314,87 @@ class Path(id: String? = null, baseClass: String? = null, job: Job, scope: Scope
...
+     */
+    fun <V> Store<List<V>>.renderEach(
+        idProvider: IdProvider<V, *>,
+        into: Tag<HTMLElement>? = null,
+        content: RenderContext.(Store<V>) -> Tag<HTMLElement>
+    ) {
+        data.renderEach(idProvider, into) { value ->
+            content(this@renderEach.sub(value, idProvider))
+        }
+    }
```
jamowei, line 1397: Is this function really needed? It's only a convenience function that may confuse!?

core/src/jsMain/kotlin/dev/fritz2/dom/html/elements.kt
```diff
@@ -1337,7 +1404,7 @@ interface TagContext : WithJob, WithScope {
      */
     fun <T> Flow<T>.asString(): Flow<String> = this.map { it.toString() }
```
jamowei, line 1405: Do we wanna keep that helper function? In my opinion this is confusion cause it appears next to `asText()` and is only a simple mapping instead of creating a mount-point.

jamowei (dismissed):
I cleaned up some things and added to comments.

chausknecht (approved):

PR #545, opened 2021-11-20
Transitions (css-animations on DOM manipulation)

jwstegemann:
This PR adds transition-support to fritz2.

You now can define a css-transition by the css classes describing the transition itself as well as the start- and endpoint like

```kotlin
val fade = Transition(
    enter = "transition-all duration-1000", 
    enterStart = "opacity-0",
    enterEnd = "opacity-100",
    leave = "transition-all ease-out duration-1000",
    leaveStart = "opacity-100",
    leaveEnd = "opacity-0"
)
```

and apply it to a tag:

```kotlin {
        div(id = "myDiv") {
            inlineStyle("margin-top: 10px; width: 200px; height: 200px; background-color: red;").   
            transition(fade)
        }
```

(Remove for release note)
Find working example code at: https://github.com/jwstegemann/fritz2-template/tree/jwstegemann/transitions

Also a new method is offered @ Tag to render context as a sibling to the current Tag:

```kotlin
  div {
    +"some content"
  }.annex {
     span { +"annex" }
  }
  div {
    +"some other content"
  }
```

renders 

```html
<div>some content>
<span>annex</span>
<div>some other content</div>
```

**Recommendation:** Do not apply transitions to tags that are also styled dynamically (`Flow` based)! This might lead to race conditions and therefore unwanted visual effects.


chausknecht (dismissed):

chausknecht (dismissed):

chausknecht (approved):
Very nice! :-)

Issue #83, opened 2020-06-12
Error with observed attributes in WebComponents


jwstegemann:
In Safari (mobile and desktop) the initial value of an observed attribute is not available on the `Flow`created with `attributeChanges`.

PR #546, opened 2021-11-21
Use a SharedFlow for Web Component attribute changes

tradeJmark:
I was running into a similar issue to what is described in Issue #83, and while investigating the codebase to see what might be causing it, I saw that you were planning to move to using a `SharedFlow` for Web Component attribtues anyway. I believe that this will solve the issue, because `SharedFlow` is a hot flow, so we wouldn't need to wait for someone to consume the flow in order to give `attributeChangedCallback` a value. Hope this helps!


jwstegemann:
Thank you very much! Never found time to look into this. Works great and also fixes the web-component-example (one of our oldest issues...).

jwstegemann (approved):

PR #553, opened 2021-11-29
Improve key event handling

chausknecht:
This PR improves the whole key event handling API with the following apsects:
- improved `Key`-class based upon a new concept of `ExtraKey`-interface, which enables constructing nice key shortcut combinations like "Strg + K" or "Shit + Alt + F" or alike.
- add some convenience methods to improve the event handling for `KeyboardEvent`s: `keys()`
- improved `Keys`-object with predefined common keys like `Tab`, `Enter` but also the extra keys like `Alt`, `Shift`, `Meta` and `Control`
- mark some special interest functions as *deprecated* in favour of using the new `keys` functions, which does the heavy work without being too specialized.

The new key API allows easily defining combinations of keys with extra keys, constructing those from a `KeyboardEvent` and prevents meaningless combinations of "real" keys:
```kotlin
// constructing a key by hand
Key("K")
Key("K", ctrl = true)

// constructing a key from a KeyboardEvent
div {
    keydowns.map { Key(it) } handledBy { /* use Key-object for further processing */ }
    //                 ^^
    //                 use KeyboardEvent to construct a Key-object with all potentially extra key states reflected!
}

// using predefined keys from Keys object
Keys.Enter // "real" Key
Keys.Alt // ExtraKey

// constructing a key with some extra key
Key("K") + Keys.Control

// defining some common combination:
val searchKey = Keys.Control + Keys.Shift + "F"
//              ^^^^^^^^^^^^
//              You can start with an extra key!
//              Appending a String to an ExtraKey will finally lead to a Key

val tabbing = setOf(Keys.Tab, Keys.Shift + Keys.Tab)

// API prevents accidently usage:
// won't compile!
Key("F") + Key("P") // cannot combine real keys!
```

As most of the time you will use keys within the handling of some KeyboardEvent, there are some new methods for leveraging some common work by the framework. Those `keys`-methods variants accept one, arbitrary or a set of `Key`s which should be handled. All events from other keys will be dropped, so that one can prevent default behaviour or stop bubbling explicitly only for the handled keys. The functions return a pair of the key and the corresponding event, so that calling `stopPropagation` or accessing other properties of the event is still possible for the client.

Example usage:
```kotlin
// within some Tag<*> or WithDomNode<*>

// do something only for some key shortcut combination
keydowns.keys(Keys.Strg + "K").map { } handledBy { /* popup some command entry component */ }

// accessing the event after the filtering:
keydowns.keys(Keys.ArrowUp, Keys.ArrowDown).mapNotNull { (key, e) ->
    e.preventDefault() // page should not scroll up or down in this case!
    when (key) {
        Keys.ArrowDown -> // create / modify something to be handled
        Keys.ArrowUp -> // create / modify something to be handled
        else -> null
    }
} handledBy { ... }
```

## Attention
This PR is based upon the Transitions branch, so the latter should be merged first!


jwstegemann (dismissed):

jamowei (changes requested):
Good idea, but I would suggest some more improvements.
- `key()` function should return a `Pair<Key, KeyboardEvent>` too (similartity)
- I would prefer same functions for `WindowListener` too

core/src/jsMain/kotlin/dev/fritz2/dom/listener.kt
```diff
@@ -162,13 +163,47 @@ fun DomListener<Event, HTMLSelectElement>.selectedText(): Flow<String> =
...
+ * Gives you the pressed key as [Key] combined with the [KeyboardEvent] filtered by exactly one key.
+ * All other events from other keys will be dropped!
+ *
+ * This is very helpful if the event bubbling should be stopped for example, as the filtering has to be done before
+ * and the bubbling should in most cases only be stopped for the handled key and for other ones!
+ *
+ * @param key the key to be handled
+ */
+fun <X : Element> DomListener<KeyboardEvent, X>.keys(key: Key) = events.map { Key(it) to it }.filter { it.first == key }
+
```
jamowei, line 199: Why you filter here? In my opinion this is not needed. Am I wrong?

jamowei (dismissed):

chausknecht:
> Good idea, but I would suggest some more improvements.
> 
>     * `key()` function should return a `Pair<Key, KeyboardEvent>` too (similartity)
This would break API though! But I agree on the consistency aspect.
As we will break lots of API for the upcoming 0.14 release anyways and this is probably the last chance to do, I would vote for this proposal!

> 
>     * I would prefer same functions for `WindowListener` too

Have you read my discussion section in the PR comment? If so, please try to argue why this would make sense :-)

jamowei (dismissed):
Nice work! Perhaps you can define some tests for the different +-operator functions ;)

chausknecht:
Unit Tests should be definitly added before merging! I will complete this later on.

chausknecht:
Closed due to Github trouble (build process is stalled!).

See PR #556 instead.

PR #555, opened 2021-12-08
fix modal close handler

metin-kale-cf:
the close handler in the modal does not always close the correct modal, it does actually only close the modal, which is on top at that moment.

For example triggerring the handler more than once also closes the modal below. When closing a modal and opening another modal simultaneously it is also possible that the wrong modal gets closed .

Instead of closing the modal on top, this pull request removes the explicit modal from the stack to fix these issues.

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/modal/component.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/modal/component.kt
@@ -3,7 +3,6 @@ package dev.fritz2.components.modal
 import dev.fritz2.binding.RootStore
 import dev.fritz2.binding.SimpleHandler
 import dev.fritz2.binding.storeOf
-import dev.fritz2.binding.watch
 import dev.fritz2.components.flexBox
 import dev.fritz2.components.foundations.*
 import dev.fritz2.dom.html.Div
@@ -20,7 +19,6 @@ import dev.fritz2.styling.staticStyle
 import dev.fritz2.styling.theme.*
 import kotlinx.browser.document
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.flow.map
 import org.w3c.dom.get
 
 /**
@@ -155,8 +153,8 @@ open class ModalComponent :
             stack + dialog
         }
 
-        val pop = handle { stack ->
-            stack.dropLast(1)
+        fun pop(dialog: ModalRenderContext) = handle { stack ->
+            stack - dialog
         }
     }
 
@@ -256,7 +254,7 @@ open class ModalComponent :
         id: String?,
         prefix: String
     ): SimpleHandler<Unit> {
-        val close = stack.pop
+        lateinit var close: SimpleHandler<Unit>
 
         val modal: ModalRenderContext = { level ->
             flexBox({
@@ -305,6 +303,8 @@ open class ModalComponent :
                 }
             }
         }
+
+        close = stack.pop(modal)
 
         return stack.push(modal)
     }


chausknecht:
> the close handler in the modal does not always close the correct modal, it does actually only close the modal, which is on top at that moment.

Which is totally correct and the desired bahaviour so far...

> For example triggerring the handler more than once also closes the modal below.

This is definitly correct. I wonder how this could ever be applied wrong? Can you elaborate, how one could (accidently) achieve this?

> When closing a modal and opening another modal simultaneously it is also possible that the wrong modal gets closed .

How is this possible? ("Simultaneously" is rather impossible in JS world, as everything gets linearized before entering the DOM rendering loop, but ok, thats more a technical wording mismatch) You mean a button that triggers two dialogs at once or alike? I can not imagine some use case where such a behaviour would make sense! Can you give a meaningful example?

> Instead of closing the modal on top, this pull request removes the explicit modal from the stack to fix these issues.

This works technically, but it hurts the core idea of a *modal*! You only ever have *one* at a time that is active and it is *always* by definition the top most one. No other can be opened at the same time, you cannot open another on the "same level". So why should a real stack not be sufficient?

BTW: Even if noone might need it, the overlay strategy might not work well with your approach of ripping some modal out of the middle of the stack. Have you checked that?

metin-kale-cf:
Yes if you think based on user interaction, it seems to be very difficult to achieve duplicate close calls or calling close from a modal, which is not on top.

I have a generic loading dialog, which wraps a task of the type (suspend () -> R) to show the dialog during the task. It shows the modal before that task starts and exists after it has been finished. So no user interaction is involved here at all. 

The task running inside might start other modals and every change in the modal stack triggers a rerender for all modals (which should actually be fixed also). That means that the code, which is waiting for the task to finish might be called twice, so that close is also called twice.
==> this was quite easy to fix with a simple variable to check, whether close was already called or not

But secondly when the loading dialog wants to close itself, it does not know if other modals have been started already and just will call "close". If another modal was started, it will close the wrong modal.

I understand the core idea of a stack, but i think enforcing it like this only creates confusion. I think a close handler of a modal should only be able to close that modal only.

chausknecht:
I can understand the problem now. But I am still not convinced that the modal approach must adapt to it, rather than to *fix* the UI process! The modal itself works (with one drawback you also have discovered, see last section of this comment) - it is the confusion of "state" within your UI processes that causes the problems!

> I have a generic loading dialog, which wraps a task of the type (suspend () -> R) to show the dialog during the task. It shows the modal before that task starts and exists after it has been finished. So no user interaction is involved here at all.

That is fine imho. And it does not cause any problems itself, right?

> But secondly when the loading dialog wants to close itself, it does not know if other modals have been started already and just will call "close". If another modal was started, it will close the wrong modal.

This is a missconception imho: The "automatic" closing simply can't be separated from the user interaction within the process timeline. Both need to be combined for deciding whether the overall process has finished and the final "close" can be called.

That said I am not sure, whether the initial approach is right. If you have some long running process, it is a *bad* idea to request user interaction within! Strive for a solution where the user input is done at the beginning and then start the long running process without any user interaction. It has so many advantages...

> ...every change in the modal stack triggers a re-render for all modals (which should actually be fixed also).

This would be a nice thing. It would require to be even more strict with just closing the outer most modal, as you would have to check within the iteration just to render the top most "added" dialog and to re-render the overlay.

metin-kale-cf:
> That said I am not sure, whether the initial approach is right. If you have some long running process, it is a _bad_ idea to request user interaction within! Strive for a solution where the user input is done at the beginning and then start the long running process without any user interaction. It has so many advantages...

Yes it is a bad idea, but that was even not the case. I was emitting informations within the task, which was listened for in another place, the task itself did not depend on that modal. 

I now managed to fix, but i still think that the current state could lead to unpredictable problems. For my problem the modals where logically connected to each other and i could control, when to close and when to start which modal, but for example if there are several tasks in an application, which might show a dialog. It could lead to undefined behaviour if the wrong modal gets closed. 

A stack like behaviour sounds very promising but in practise i think it will cause more problems. I actually can not think of any sideeffects or problems which occur by breaking the stack and closing modals, which are not on the top. But as i said, i solved my problem, its up to the fritz2-Team to make the decision.

chausknecht:
We will rework whole modal approach - so I close this issue here, because we will keep this problem in mind for new solution :-)

PR #556, opened 2021-12-20
Improve Key API and keys event handling

chausknecht:
This PR improves the whole key event handling API by the following aspects:
* improved `Key`-class based upon a new concept of `Modifier`-interface, which enables constructing nice key shortcut combinations like "Strg + K" or "Shift + Alt + F" or alike.
* add some convenience methods to improve the event handling for `KeyboardEvent`s: `keys()`
* improved `Keys`-object with predefined common keys like `Tab`, `Enter` but also the modifier keys like `Alt`, `Shift`, `Meta` and `Control`
* mark some special interest functions as _deprecated_ in favour of using the new `keys` functions, which does the heavy work without being too specialized.

The new key API allows easily defining combinations of keys with modifier keys, constructing those from a `KeyboardEvent` and prevents meaningless combinations of "real" keys:

```kotlin
// constructing a key by hand
Key("K") // will convert "K" automatically to lower case -> Key(name = "k", ctrl = false, alt = false, shift = false, meta = false)
// same as
Key("k")
// set modifier states
Key("K", ctrl = true) // Key(name = "k", ctrl = true, alt = false, shift = false, meta = false)

// constructing a key from a KeyboardEvent
div {
    keydowns.map { Key(it) } handledBy { /* use Key-object for further processing */ }
    //                 ^^
    //                 use KeyboardEvent to construct a Key-object with all potentially modifier key states reflected!
}

// using predefined keys from Keys object
Keys.Enter // "real" Key
Keys.Alt // ModifierKey -> needs to be combined with a "real" key in order to use it for further processing
// the same but more cumbersome and open to typos
Key("Enter")
// not the same (!)
Key("Alt") // -> Key(name = "Alt", ..., alt = false)
Keys.Alt // -> ModifierKey-object with alt = true property!

// constructing a key with some modifier key
Key("K") + Keys.Control
// same result, but much more readable the other way round:
Keys.Control + "K"

// defining some common combination:
val searchKey = Keys.Control + Keys.Shift + "F"
//              ^^^^^^^^^^^^
//              You can start with a modifier key!
//              Appending a String to an ModifierKey will finally lead to a Key

val tabbing = setOf(Keys.Tab, Keys.Shift + Keys.Tab)

// API prevents accidently usage:
// won't compile!
Key("F") + Key("P") // cannot combine real keys!
```

As most of the time you will use keys within the handling of some `KeyboardEvent`, there are some new methods for leveraging some common work by the framework. Those `keys`-methods variants accept one, arbitrary or a set of `Key`s which should be handled. All events from other keys will be ignored, so that one can prevent default behaviour or stop bubbling explicitly only for the handled keys. The functions return a pair of the key and the corresponding event, so that calling `stopPropagation` or accessing other properties of the event is still possible for the client.

Example usage:
```kotlin
// within some Tag<*or WithDomNode<*>

// do something only for some key shortcut combination
keydowns.keys(Keys.Strg + "K").map { } handledBy { /* popup some command entry component */ }

// accessing the event after the filtering:
keydowns.keys(Keys.ArrowUp, Keys.ArrowDown).mapNotNull { (key, e) ->
    e.preventDefault() // page should not scroll up or down in this case!
    when (key) {
        Keys.ArrowDown -> // create / modify something to be handled
        Keys.ArrowUp -> // create / modify something to be handled
        else -> null
    }
} handledBy { /* ... */ }
```

### DIFF
--- a/components/src/jsMain/kotlin/dev/fritz2/components/popover.kt
+++ b/components/src/jsMain/kotlin/dev/fritz2/components/popover.kt
@@ -9,7 +9,7 @@ import dev.fritz2.dom.Window
 import dev.fritz2.dom.html.Key
 import dev.fritz2.dom.html.Keys
 import dev.fritz2.dom.html.RenderContext
-import dev.fritz2.dom.key
+import dev.fritz2.dom.keys
 import dev.fritz2.styling.*
 import dev.fritz2.styling.params.BasicParams
 import dev.fritz2.styling.params.BoxParams
@@ -20,7 +20,6 @@ import dev.fritz2.styling.theme.PopoverSizes
 import dev.fritz2.styling.theme.Theme
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.debounce
-import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.flowOf
 import org.w3c.dom.HTMLElement
 
@@ -167,7 +166,7 @@ open class PopoverComponent : Component<Unit>,
         context.apply {
 
             if (this@PopoverComponent.closeOnEscape.value) {
-                Window.keyups.key().filter { it == Keys.Escape } handledBy this@PopoverComponent.visible.closeOnKey
+                Window.keyups.keys(Keys.Escape) handledBy this@PopoverComponent.visible.closeOnKey
             }
 
             div(staticCss.name, id) {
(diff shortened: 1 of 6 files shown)


chausknecht:
Question: Would it be nice from the ``Key`` ctor to call a ``lowercase`` on all Strings? A client might be trapped by the current strict behaviour to define some ``Keys.Control + "F"`` with an uppercase "F". But this will never ever be recognized by the `keys` functions as the DOM API only emits lowercase characters... to achieve a reaction to a "STRG + K" combination the key must be constructed with a lower case character right now: ``Keys.Control + "f"``.

On the other hand it would be at least one call on top and also would probably need some sort of filter to only apply the lowercase to one length strings (as the predefined all starts with upper case!)

chausknecht:
After dicussing the issue with lower- vs upper-case keys, we have decided to look for a rational implementation that will detect all none special keys and transform them to lower-cases. This way the API is much more pleasent as it would make no difference if one writes ``Keys.Control + "K"`` of ``Keys.Control + "k"`` or even ``Key("K")`` vs ``Key("k")``.

jamowei (approved):

PR #558, opened 2021-12-28
Added `Middleware` support to html API + pre-build stateless `Authentication` middleware

jamowei:
In the fritz2 `http` API you can now add `Middleware`s which have the following definition:
```kotlin
interface Middleware {
    suspend fun enrichRequest(request: Request): Request
    suspend fun handleResponse(response: Response): Response
}
```
You can add an `Middleware` to all your http calls by using the new `use(middleware: Middleware)` function.
```kotlin
val logging = object : Middleware {
    override suspend fun enrichRequest(request: Request): Request {
        console.log("doing request: $request")
        return request
    }

    override suspend fun handleResponse(response: Response): Response {
        console.log("getting response: $response")
        return response
    }
}

val myAPI = http("/myAPI").use(logging)
...
```

You can add multiple `Middleware`s in one row by `.use(mw1, mw2, mw3)`. The `enrichRequest` functions will be called from left to right (mw1,mw2,mw3), the `handleResponse` functions back from right to left (mw3, mw2, mw1). You can stop the processing of a `Response` by `Middleware`s further down the chain by `return response.stopPropagation()`.

Also we build a pre-implemented `Authentication` middleware to enrich all request with authentication information and handle bad authentication responses in general.
```kotlin
abstract class Authentication<P> : Middleware {

    // list of status codes forcing authentication
    open val statusCodesEnforcingAuthentication: List<Int> = listOf(401, 403)

    // add your authentication information to all your request (e.g. append header value) 
    abstract fun addAuthentication(request: Request, principal: P?): Request

    // start your authentication process (e.g. open up a login modal)
    abstract fun authenticate()

    // manually start your authentication process before response with forcing status code returns
    fun start()
    
    // complete the authentication process by setting a new principal
    fun complete(principal: P)
    
    // clear current authentication information (principal)
    fun clear()

    val authenticated: Flow<Boolean>
   
    val principal: Flow<P?>
}
```


core/src/jsMain/kotlin/dev/fritz2/remote/auth.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    private val principalStore = storeOf<P?>(null)
+
+    private var state: CompletableDeferred<P>? = null
+
+    /**
+     * List of HTTP-Status-Codes forcing an authentication.
+     * Defaults are 401 (unauthorized) and 403 (forbidden).
+     */
+    open val statusCodesEnforcingAuthentication: List<Int> = listOf(401, 403)
```
chausknecht, line 32: As the only operation applied onto this is looking up a specific code, this should be a `Set` and not a `List`. I also cannot imagine some use case the sequential order will ever be relevant.

core/src/jsMain/kotlin/dev/fritz2/remote/http.kt
```diff
@@ -6,229 +6,306 @@ import org.w3c.fetch.*
...
+     */
+    suspend fun json() = response.json().await()
+
+}
+
+/**
+ * interface to do interceptions at http calls.
+ * It can modify each request and handle all responses in a specified way.
+ */
+interface Interceptor {
```
chausknecht, line 81: Lost of Other frameworks call this "middleware". I would suggest adapting to the common wording though.

core/src/jsMain/kotlin/dev/fritz2/remote/http.kt
```diff
@@ -6,229 +6,306 @@ import org.w3c.fetch.*
...
 
 /**
- * creates a new [Request]
- *
- * @param baseUrl the common base of all urls that you want to call using the template
+ * Represents the common fields and attributes of an HTTP response.
+ * It contains also the original [Request] which was made to get this [Response].
  */
-fun http(baseUrl: String = "") = Request(baseUrl = baseUrl)
+open class Response(
```
chausknecht, line 33: The sequence of appearance has changed. I am not sure whether this is intentional ot not. Imho the sequence in its current form does not follow any paradigm. 

If we want to present the base concepts first and subsequently add dependent concepts, the order would be:
- Response
- FetchException
- Request
- Interceptor
- fun http

core/src/jsMain/kotlin/dev/fritz2/remote/auth.kt
jamowei: changed

core/src/jsMain/kotlin/dev/fritz2/remote/http.kt
```diff
@@ -6,229 +6,306 @@ import org.w3c.fetch.*
...
-                val redo = authentication.enrichRequest(this).buildInit(init.method!!)
-                response = browserWindow.fetch(url, redo).await()
-            }
-        }
+     * @throws FetchException when request failed
+     */
+    suspend fun execute(): Response {
+
+        var request = this
+        for(interceptor in interceptors) request = interceptor.enrichRequest(request)
```
chausknecht, line 159: This approach does apply all interceptors in any case. The interceptor itself has no possibility to abort the upcoming interceptors. I am not sure whether this feature is needed (often), but other frameworks rely on the "chain of responsibility" pattern instead (servelt API for example). It might be possible though, that the latter simply is not needed at client side. But it might be worth thinking about before we finalize this API!

One can quite easily realize this of course by implementing such a chaining mechanism within a single interceptor and then manage all "real" interceptors from within. But that would bypass our API and would feel quite cumbersome.

chausknecht, line 159: How about using fold?
````kotlin
var request = interceptors.fold(this) { request, interceptor -> interceptor.enrichRequest(request) }
````

File: core/src/jsMain/kotlin/dev/fritz2/remote/http.kt [hunk 5]
```diff
@@ -6,229 +6,306 @@ import org.w3c.fetch.*
...
+
+        val init = request.buildInit()
+
+        var response = Response(browserWindow.fetch(url, init).await(), request)
+        for(interceptor in interceptors) response = interceptor.handleResponse(response)
```
chausknecht, line 164: like above
````kotlin
        var response =
            interceptors.fold(Response(browserWindow.fetch(url, init).await(), request)) { response, interceptor ->
                interceptor.handleResponse(response)
            }
````

jwstegemann (approved):

jwstegemann:
I have changed / implemented everything we discussed.

PR #561, opened 2022-01-04
Improved Validation API

jamowei:
First, we added a `MutableList` receiver to the `Validator`.`validate()` method for directly adding messages to the list. So you can call the `add()` function directly to add your message to the list:
```kotlin
object CarValidator : Validator<Car, Message, Unit>() {
    override fun MutableList<Message>.validate(inspector: Inspector<Car>, metadata: Unit) {
        if (inspector.data.name.isBlank()) add(carNameIsBlank)

        if (inspector.data.color.r < 0 || inspector.data.color.g < 0 || inspector.data.color.b < 0)
            add(colorValuesAreToLow)

        if (inspector.data.color.r > 255 || inspector.data.color.g > 255 || inspector.data.color.b > 255)
            add(colorValuesAreToHigh)
    }
}
```
For API-streamlining reasons, we renamed the global `inspect()` function to `inspectorOf` (analog to `storeOf` function).

Next improvement is that we now have a new marker interface `ValidatingStore` for stores which include a `Validator` for their model. This could be type-checked on stores to see if a validator with validation messages is available for a store or its parent-store. Also, there is a convenience method inside to enable continuous validation when your data gets updated, called `enableContinuousValidation(medata: T)`.
```kotlin
object CarStore : RootStore<Car>(Car()), ValidatingStore<Car, Message, Unit> {
    override val validator: Validator<Car, Message, Unit> = CarValidator

    init {
        // validation message list changes automatically when store data gets updated
        enableContinuousValidation(Unit)
    }
}
```
To simplify the creation of a `ValidatingStore` for less complex data models, we created a new global `storeOf` convenience function which gets a lambda expression for validation:
```kotlin
val store = storeOf<String, Message, Unit>("foo") { inspector, _ ->
    if(inspector.data.isBlank()) add(Message("input is blank"))
}.apply { enableContinuousValidation(Unit) }
```


jamowei:
resolved by #570

PR #570, opened 2022-01-20
New Validation approach

jamowei:
First, you need to specify a `Validation` for your data-model. Therefore, you can use one of the two new global convenience functions:
```kotlin
// creates a Validation for data-model D with metadata T and validation-messages of M
fun <D, T, M> validation(validate: MutableList<M>.(Inspector<D>, T?) -> Unit): Validation<D, T, M>

// creates a Validation for data-model D and validation-messages of M
fun <D, M> validation(validate: MutableList<M>.(Inspector<D>) -> Unit): Validation<D, Unit, M>
```
These functions are available in the `commonMain` source set, so you can create your `Validation` object right next to your data classes to keep them together. Example:
```kotlin
@Lenses
data class Person(
    val name: String = "",
    val height: Double = 0.0,
) {
    companion object {
        val validation = validation<Person, String> { inspector ->
            if(inspector.data.name.isBlank()) add("Please give the person a name.")
            if(inspector.data.height < 1) add("Please give the person a correct height.")
        }
    }
}
```
Then you can call your `Validation` everywhere (e.g. JVM- or JS-site) to get a list of messages which shows if your model is valid or not. We recommend extending your validation messages from the `ValidationMessage` interface. Then your validation message type must implement the `path` which is important for matching your message to the corresponding attribute of your data-model and the `isError` value which is needed to know when your model is valid or not:
```kotlin
data class MyMessage(override val path: String, val text: String) : ValidationMessage {
    override val isError: Boolean = text.startsWith("Error")
}

// change your Validation to use your own validation message
val validation = validation<Person, MyMessage> { inspector ->
    val name = inspector.sub(Person.name())
    if (name.data.isBlank())
        add(MyMessage(name.path, "Error: Please give the person a name."))

    val height = inspector.sub(Person.height())
    if (height.data < 1)
        add(MyMessage(height.path, "Error: Please give the person a correct height."))
}

// then you can use the valid attribute to check if your validation result is valid or not
val messages: List<MyMessage> = Person.validation(Person())
messages.valid  // == false
```
# New ValidatingStore
We introduce a new type of `Store` which we call `ValidatingStore`. This `Store` has the same properties as a `RootStore` and additionally a `Validation` which it uses to validate the stored model. With these additional information you get next to your known `data` flow also a `messages` flow which you can use to render out your list of validation messages. Furthermore, you can decide on store creation if you want to automatically validate your model after an update to your store takes place. Then you have to set the `validateAfterUpdate` flag and the flow of messages gets accordingly updated. Otherwise, you can call the `validate(data: D, metadata: T? = null)` function inside your own handlers to update the list of messages by your own. For cases, you want to reset your validation state to an empty list or to a specific list of messages, you can use the `resetMessages(messages: List<M> = emptyList())` function. Example:
```kotlin
// create your ValidatingStore
val myValidationStore = object : ValidatingStore<Person, Unit, MyMessage>(
    Person(), Person.validation, validateAfterUpdate = true, id = "myPerson"
) {
    // check if your model is valid and then save
    val save = handle { person ->
        if(validate(person).valid) {
            localStorage.setItem(person.name, person.height.toString())
            Person()
        } else person
    }
    //...
}

render {
    //...
    input {
        val name = myValidationStore.sub(Person.name())
        value(name.data)
        changes.values() handledBy name.update
    }
    //...
    button {
        +"Save"
        clicks handledBy myValidationStore.save
    }

    // render out the messages
    myValidationStore.messages.renderEach(MyMessage::path) {
        p {
            +it.text
            if(it.isError) inlineStyle("color: red")
        }
    }
}
```


jamowei:
Opened up a ticket at Jetbrains Youtrack: https://youtrack.jetbrains.com/issue/KT-50897

core/src/jsMain/kotlin/dev/fritz2/binding/store.kt
```diff
@@ -258,16 +260,86 @@ open class RootStore<T>(
...
      * a simple [SimpleHandler] that just takes the given action-value as the new value for the [Store].
      */
-    override val update = this.handle<T> { _, newValue -> newValue }
+    override val update = this.handle<D> { _, newValue -> newValue }
 }
 
 /**
- * convenience method to create a simple [RootStore] without any handlers, etc.
+ * A [ValidatingStore] is a [RootStore] which also contains a [Validation] for its model.
  *
```
chausknecht, line 268: This store is intentionally configured to validate the data on each update, that is why the [validateAfterUpdate] parameter is set to ``true`` by default.
There might be special situations where it is reasonable to disable this behaviour and to prefer applying the validation individually within custom handlers, for example if a model should only be validated after the user has completed his input or if meta-data is needed for the validation process. Then be aware of the fact, that the call of the [validate] function actually updates the [messages] [Flow] already. If the new data is not passed to the store's [data] flow afterwards, the messages are out of sync with the actual data state! This could lead to false assumptions and might produce hard to detect bugs in your application context. 

So for scenarios where the model in a store should always be valid, do not use this store implementation! Instead, strive for a custom store implementation where you can handle the validation completely by yourself. As the validation is simply a call on the thin [Validation] wrapper, it is really easy to manage manually.

Using this store you should always insert exactly the data, that leads to the current validation messages!

core/src/jsMain/kotlin/dev/fritz2/binding/store.kt
```diff
@@ -258,16 +260,86 @@ open class RootStore<T>(
...
+    private val validation: Validation<D, T, M>,
+    val validateAfterUpdate: Boolean = true,
+    override val id: String = Id.next()
+) : RootStore<D>(initialData, id) {
+
+    private val validationMessages: MutableStateFlow<List<M>> = MutableStateFlow(emptyList())
+
+    /**
+     * [Flow] of the [List] of validation-messages.
+     * Use this [Flow] to render out the validation-messages.
```
chausknecht, line 285: ... and to detect the valid state of the current [data] [Flow].

core/src/jsMain/kotlin/dev/fritz2/binding/store.kt
```diff
@@ -258,16 +260,86 @@ open class RootStore<T>(
...
+     */
+    val messages: Flow<List<M>> = validationMessages.asStateFlow()
+
+    /**
+     * Resets the validation result.
+     *
```
chausknecht, line 291: Beware that cleaning the messages should not be done, if the [data] [Flow] remains in an invalid state. Please refer to the class's description for details about the need for a sound data and messages state.

core/src/jsMain/kotlin/dev/fritz2/binding/store.kt
```diff
@@ -258,16 +260,86 @@ open class RootStore<T>(
...
+ * @param id id of this store. Ids of [SubStore]s will be concatenated.
+ */
+fun <D> storeOf(initialData: D, id: String = Id.next()) = RootStore(initialData, id)
+
+/**
+ * Convenience function to create a simple [ValidatingStore] without any handlers, etc.
+ * The created [Store] validates its model after every update automatically.
+ *
+ * @param initialData first current value of this [Store]
+ * @param validation [Validation] function to use at the data on this [Store].
```
chausknecht, line 338: * @param validation [Validation] instance to use at the data on this [Store].

core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
```diff
@@ -1,96 +1,89 @@
...
-        state.value = messages
-        return messages.none(ValidationMessage::isError)
-    }
-
-    /**
-     * A [Flow] representing the current state of the model (valid or not).
-     */
-    val isValid: Flow<Boolean> by lazy { data.map { list -> list.none(ValidationMessage::isError) } }
-}
+///**
```
chausknecht, line 3: Should be deleted!

core/src/jvmMain/kotlin/dev/fritz2/validation/validation.kt
```diff
@@ -1,64 +1,61 @@
...
-     *
-     * @param data model to validate
-     * @param metadata extra information for the validation process
-     * @return a [Boolean] for using in if conditions
-     */
-    fun isValid(data: D, metadata: T): Boolean = validate(inspect(data), metadata)
-        .also { state = it }.none(ValidationMessage::isError)
-}
\ No newline at end of file
+///**
```
chausknecht, line 3: Should be deleted as well!

chausknecht (changes requested):
Some proposals for the API docs.

Besides this looks really good! I hardly can wait to use this :-)

I have some proposals for the release notes too. We should talk about those afterwards.

chausknecht (dismissed):

chausknecht (approved):

PR #571, opened 2022-01-21
`Authentication.getPrincipal()` function returns the principal also when pre-setting it before any process is started

### LINKED ISSUE #577: Authentication Middleware - getCurrentPrincipal
it would be helpful to provide a "getCurrentPrincipal" function that is not suspendable and returns the current object from the store

fun getCurrentPrincipal(): principalStore.value

jamowei:
Fixes #577

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/remote/auth.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/remote/auth.kt
@@ -1,8 +1,9 @@
 package dev.fritz2.remote
 
-import dev.fritz2.binding.storeOf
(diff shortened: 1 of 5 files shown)


jwstegemann:
Should we rename 'getPrincipal' to 'current' since this is what we generally offer to access the current value in a Store, Router, etc.?

jamowei:
> Should we rename 'getPrincipal' to 'current' since this is what we generally offer to access the current value in a Store, Router, etc.?

Yeah, I change it.

chausknecht (approved):

PR #576, opened 2022-01-27
Add headless toasts

haukesomm:
This PR adds headless toasts as well as a simple demo implementation.


haukesomm:
Abandoned in favor of #660

PR #579, opened 2022-01-28
Fixed exception handling in Handlers

jwstegemann:
To make sure, that the coroutine of a `Store` or `RenderContext` is not cancelled by an uncaught exception, we changed the way of exception handling in fritz2:

You cannot provide a lambda specifying, how to handle uncaught exceptions, to a handler anymore. Just use `try ... catch` 
- in your handler's code, 
- in any intermediate operation on the `Flow` or  
- by using `.catch()` on the `Flow` 

to deal with an exception the way you want.

An exception you did not catch earlier (by `try ... catch` or `.catch()` on a `Flow`) will be caught by fritz2 for any type of handler, including ad hoc handlers.

By default, the exception will be just logged to the console. You can change this behavior be overriding the `errorHandler` function of your `Store`, `RenderContext`, etc.

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/binding/store.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/binding/store.kt
@@ -21,37 +21,9 @@ import kotlinx.coroutines.sync.withLock
 typealias Update<T> = suspend (T) -> T
 
 /**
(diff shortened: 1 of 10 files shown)


jamowei (approved):
Nice work!

PR #580, opened 2022-01-30
Transitions controlled by Flow

jwstegemann:
In addition to transitions that are run, when a `Tag` is added to or removed from the DOM, you can now also apply a `Transition` to a mounted `Tag` whenever a new value appears on a `Flow[Boolean]`:

```kotlin
tag.transition(someFlowOfBoolean,
    "transition duration-100 ease-ease-out",
    "opacity-0 scale-95",
    "opacity-100 scale-100",
    "transition duration-100 ease-ease-out",
    "opacity-100 scale-100",
    "opacity-0 scale-95"
)
```

The enter-transition will be executed when `true` appears on the [Flow]
The leave-transition will be executed when `false` appears on the [Flow]

We also added a suspending extension-function `WithDomNode<*>.waitForAnimation`, that allows you to wait for running transitions to be finished. This is for example useful, if you want to make a parent element invisible only after a the fading-out of it's children is done.


chausknecht (approved):
Looks nice!

chausknecht (approved):

PR #589, opened 2022-02-07
Improve Tailwind setup

jamowei:
* use only relative paths
* less config files
* no need for `require()` in Kotlin code anymore


jwstegemann (approved):
Perfect. A huge improvement...

jwstegemann (approved):

PR #593, opened 2022-02-14
Move `storeOf` function out of `validation` package

jamowei:


chausknecht (commented):
I think this is not right!

You artificially create a cyclic dependency between store.kt and validation.kt. The `ValidatingStore` is some special store, so it is *dependent* from the store package.

Imho this automatically defines the location of the *specific* `storeOf` function too. Right were it was ;-)

So I would reject this whole PR.

jamowei:
> I think this is not right!
> 
> You artificially create a cyclic dependency between store.kt and validation.kt. The `ValidatingStore` is some special store, so it is _dependent_ from the store package.
> 
> Imho this automatically defines the location of the _specific_ `storeOf` function too. Right were it was ;-)
> 
> So I would reject this whole PR.

Valid point. But in my opinion it is more important for the user, that overloading of the global `storeOf` function works. Otherwise you have to know to import an extra `validation` package to get the right autocompletion. That's the reason why I moved it back. Then it's much easier to see, that there is an extra `storeOf` function with validation parameter! 😉

chausknecht:
So we need some 3rd opinion on that one :-D

jwstegemann:
Code completion seems to work for me as it is (without import):

Bildschirmfoto 2022-02-14 um 20 39 52

jamowei:
> Code completion seems to work for me as it is (without import):
> 
> Bildschirmfoto 2022-02-14 um 20 39 52

Changes are already in master, that's why you can't see the difference!

jamowei:
You have to import one of the `storeOf` functions fist (`dev.fritz2.validation` or `dev.fritz2.binding`). Then IntelliJ IDEA gets confused when you want to see the available parameters (dialog stays not open):
image

With both function in the same package, it looks common:
image

I think it is common to use both functions in one file, and the IDE support is better with having both functions in the same package. Hopefully my point of view gets clearer now.

jwstegemann:
No, in a blank template-project auto-completion without any import works great:

Bildschirmfoto 2022-02-15 um 17 03 15

PR #596, opened 2022-02-17
New package structure

jamowei:
Following changes takes place:
*  one `core` package for all core features
* remove sub-package `html`
* rename `repositories` package to `repository`
* rename `lenses` package to `lens`
* new `foundation.kt` in `core` package for global functions
* move `scope.kt` and `job.kt` to `core` package
* put `tag.kt`, `htmltag.kt` and `svgtag.kt` into new `tags.kt` file
* align test files to new package structure
* align commonMain packages to jsMain packages


jwstegemann:
Looks much cleaner than before, but why is validation part of core and history, tracking, repos, etc. not?

chausknecht (approved):

PR #600, opened 2022-02-21
Fixes and improvements for new webpage

jamowei:


www/src/index.njk
```diff
@@ -28,11 +28,11 @@ layout: layouts/landing.njk
               </div>
             </div>
             <div class="hidden space-x-8 md:flex md:ml-10">
-              <a href="/docs" class="text-base font-medium text-white hover:text-gray-300">Documentation</a>
+              <a href="/docs" target="_blank" class="text-base font-medium text-white hover:text-gray-300">Documentation</a>
```
jwstegemann, line 31: Why should anyone want to open a new tab or window? This main-page is just a landing-page and you can reach any other page from the "relevant" sub-pages

jamowei: Ok, open tab only when clicking on `API` now, because there is no link to get back to the landing page.

jwstegemann (approved):

PR #617, opened 2022-03-02
Headless DataCollection

jwstegemann:
implement a headless component for collections of data like data-tables, data-lists, etc. including sorting, filtering, keyboard-navigation, selection


chausknecht (approved):
Best Headless Data Collection on Earth so far!

PR #637, opened 2022-04-12
Set up initial playwright infrastructure

chausknecht:
- add playwright to demo project
- add first test-suite (listbox)

There is currently no integration into the build process, besides a github task, if headless-demo modul is pushed to master. We need to discuss, whether we need more automatic integration in the future or not.

As we will not change constantly the behaviour of headless components, it might be sufficient, if the developer and the reviewer start the tests manually on development or during review process locally.


jamowei (approved):
Nice work!

PR #638, opened 2022-04-14
Fix functionality of `attr()` functions

jamowei:
Streamline internal behavior of the `attr()` functions by using it each other. This also fixes the issues that it is not possible to remove an attribute by using `null` as value.


chausknecht:
> This also fixes the issues that it is not possible to remove an attribute by using null as value.

IIrc this has been possible - ist this a regression issue?

Either way, we should consider some unit test(s), that explicitly test this behaviour.

PR #639, opened 2022-04-14
Updating text area demo using github theme

jamowei:
My first try. Please take a look ;)


jamowei:
Not needed anymore. We use our own theme.

Issue #636, opened 2022-04-05
address tight coupling with ksp & kotlin version


jillesvangurp:
**Is your feature request related to a problem? Please describe.**

I just tried to update to Kotlin 1.6.20, which was just released. It failed because the ksp version has to match the kotlin version used in the lenses jar.

**Describe the solution you'd like**

Would it be possible to relax this so patch releases of Kotlin are not a problem? Alternatively, can you release a new version against 1.6.20?

**Describe alternatives you've considered**

I tried manually adding the ksp dependency with the later version but that doesn't seem to work.

**Additional context**

chausknecht:
If the KSP version of our plugin has to match the Kotlin version of the application I cannot imagine a decoupled solution. Although it feels quite odd to me, that there is such a strong coupling to be honest.

I do not know, if someone from our team has already tried to use Kotlin 1.6.20 with fritz2 already.

But I am quite sure the fritz2-1.0-RC1 release is quite close and that will be based upon the current Kotlon version, if we do not encounter any problems.

I can only encourage you, to build fritz2 locally on your own with the adapted versions! If that works, why not create a PR? It is really not that hard :-)

jillesvangurp:
Yeah that's what I suspected. I don't really have a good development environment for Fritz2. I should fix that probably but it will take me some time. I can upgrade the dependencies of course and create a PR. 

Looking forward to the release candidate btw.

jillesvangurp:
@chausknecht I created a PR that does the upgrade to 1.6.21. Sorry this took so long.

I would recommend using the refreshVersions plugin. That's a bit bigger refactor of the build but it's a great plugin for staying on top of versions of dependencies.

I can do that as well if you need me to. 

Probably a lot of the npm dependencies need updating as well.

PR #644, opened 2022-05-11
Add UI Tests for Headless Components with Playwright

bfongank:
This PR adds UI tests for the headless components by testing the demo examples.
All currently available components are tested as far as functionalities are offered by the demo examples.

PR #646, opened 2022-05-31
upgrade to kotlin 1.7 & other dependency updates

jillesvangurp:
upgrades to kotlin 1.6.21

fix ksp related things in the build (a few things changed in ksp)

Tests seem to pass.


jillesvangurp:
fixes #636

chausknecht:
As 1.7 is out, we should upgrade directly to 1.7.

I would propose to wait for PR #644 to be accomplished and then try to update this PR towards 1.7. Armed with the unit- and ui-tests we should be quite save to do that move now.

chausknecht:
@jillesvangurp Thank you! We will investigate this PR very soon. Right now I focus on the UI tests in order to have them in place before upgrading to 1.7. Stay tuned for questions or the finished RC1 :-)

chausknecht:
You are right. Bu as you have stated already, that ksp has introduced some changes, I believe we will be safer to do the "major" move to 1.7 directly and investigate carefully, what we must change for out template projects and what we should document as migration help.

We are very close to the RC1 release, so even such minimal looking aspects could disturb the release.

But your PR will not be forgotten! :-)

jwstegemann:
I just merged our 1.0-RC1 master into this branch, set coroutines-dependency to 1.6.3 (which fixes the error that @jillesvangurp mentioned above) and added a workaround for a broken webpack-cli-dependency with kotlin 1.7.
Works for me like a charm (recognizable performance gains at compile- and runtime).

jwstegemann:
Would be happy about some more tests (especially of the build on some other operating-systems and hardware)

jillesvangurp:
I'll give it a try early next week.

jamowei:
Successfully tested this PR under Linux/Ubuntu:
- [x] ./gradlew build
- [x] playwright tests
👍🏻

jamowei (approved):

jwstegemann:
Just added to new Branches to the template-projects to work with this version:

https://github.com/jwstegemann/fritz2-template/pull/14
https://github.com/jwstegemann/fritz2-tailwind-template/pull/2

PR #660, opened 2022-07-06
Headless Toast component

haukesomm:
This PR adds a headless Toast component as well as the respective demo page.


haukesomm:
Closed as there is still more work to do before opening a PR makes sense.

PR #688, opened 2022-09-15
Add default z-Index for PopUpPanel

### LINKED ISSUE #670: Set Z-Index on appropriate element inside PopUpPanel
A default z-Index should be set for all components using PopUpPanel, in order to bring its content to the front.
It should be preferably made to be exposed by the using headless component in order to enable the user to overrule the default value in special circumstances.

chausknecht:
I order to render the floating content of a PopUpPanel based component, we have defined a static z-Index with value ``30`` for the popper div, which can't be styled in a safe and reliable way from the outside.

This value corresponds to the median value of the predefined [tailwindcss](https://tailwindcss.com/docs/z-index) z-indices. 

fixes #670


jwstegemann:
I am pretty sure that it will be necessary to allow the user to set the z-index (because he uses a css-lib that has a z-index-concept like bootstrap)

jwstegemann (approved):

PR #691, opened 2022-09-16
Repair the ksp config

chausknecht:
Lenses generation was not possible due to oversimplification of the ksp config section in the build.gradle.kt. In order to preserve regressions on this topic, the ``@Lenses`` annotation were enabled again for our models in the demo, even though the generated lenses are currently not used.

Not worth mentioning in the release note!


jwstegemann:
Removed ksp-config from examples-demo module since it was broken there, too and is not needed.

jwstegemann (approved):

PR #718, opened 2022-11-03
Remove Repositories from Core

chausknecht:
TODO


chausknecht:
The very last aspect ist still open:
> "~~move code to dedicated example projects~~ and probably adapt or simplify the implementation"

I do not have the time to finish this soon probably.

chausknecht:
The removal is completed now.

I only encountered some questionable logging inside the todomvc example. As this is made using guidelines from the [well known project](https://todomvc.com/), those might intentionally remain. This is why I have not removed them already.

It would be good to know, if we can clean up the code there or not.

jwstegemann (approved):

PR #719, opened 2022-11-03
Enable `Event` capturing

### LINKED ISSUE #552: Enable Event Capturing supplementary to Event Bubbling (Default and only Method so far)
### Motivation
Currently fritz2 offers only event bubbling by default.
In order to accomplish consistent abstraction of DOM possibilities in fritz2, it would be rational to add also event capturing into the core.

- https://www.quirksmode.org/js/events_order.html
- https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#event_bubbling_and_capture

### Proposal
```kotlin
interface WithEvents<out T : Element> : WithDomNode<T> {
    // NEW:
    // extend signature to handle capturing
    private inline fun <reified E : Event> subscribe(type: EventType<E>, capture: Boolean = false): DomListener<E, T> = DomListener(
        callbackFlow {
            val listener: (Event) -> Unit = {
                if (it is E) {
                    trySend(it.unsafeCast<E>())
                } else {
                    console.error("Unexpected type while listening for `${type.name}` events")
                }
            }
            // NEW:
            // determine mode by param
            domNode.addEventListener(type.name, listener, capture)

            awaitClose { domNode.removeEventListener(type.name, listener) }
        })

    // NEW:
    // no name clashing because of parameter
    // Also avoiding boolean params is still valuable, it is ok in Kotlin because of named parameters!
    fun aborts(capture: Boolean) = subscribe(Events.abort, capture)

    val aborts
        get() = subscribe(Events.abort)
    // ...
}
```
This would allow the following call:
```kotlin
// within some Tag<*> / WithDomNode<*>
clicks(capture = true) handledBy { ... }
```
As this is also implemented as extension to the existing, property based API for "bubbling", this couls be realized without any API breaking.

In order to avoid missuse of the boolean parameter, it would be nice to handle the choice more explicitly:
```kotlin
// proposal from above, missused call:
clicks(true) handledBy { ... } // WTH is true? -> see https://docs.microsoft.com/en-us/cpp/mfc/reference/cwnd-class?view=msvc-170#updatedata for epic design fail!

// within some Tag<*> / WithDomNode<*>
captured(clicks) handledBy { ... }
// or
clicks.captured() handledBy { ... }
```

I have doubts that the latter (or another approach) would be realizable without massive API breaking though :-(

jamowei:
It is now possible to listen on events in capture-phase (suffixed by `Captured`):

```kotlin
render {
    div {
        clicksCaptured handledBy store.save
    }
}
```

For this we added new options to the `subscribe()` function which gives you a `Listener` for your event:

```kotlin
subscribe<Event>(name: String, capture: Boolean, init: Event.() -> Unit)
```

Using the `init`-lambda you can make settings to the captured event that have to be applied immediately.

We also fixed a bug when using `stopPropagation()` on a `Listener` which sometime did not work as expected.

fixes #552

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/core/events.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/events.kt
@@ -4,10 +4,7 @@ package dev.fritz2.core
 
 import kotlinx.browser.window
 import kotlinx.coroutines.MainScope
-import kotlinx.coroutines.channels.awaitClose
-import kotlinx.coroutines.flow.SharingStarted
-import kotlinx.coroutines.flow.callbackFlow
-import kotlinx.coroutines.flow.shareIn
+import kotlinx.coroutines.flow.*
 import org.w3c.dom.*
 import org.w3c.dom.Window
 import org.w3c.dom.clipboard.ClipboardEvent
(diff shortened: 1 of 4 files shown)


chausknecht (approved):
Looks very good!

The tests are running and the code looks clean.

PR #721, opened 2022-11-04
Fix CSS-Example and text

EvaBS:
- className was used to apply CSS to a tag which does not work, it needs to be inlineStyle, 
- Fixed some text.


chausknecht:
I have reverted the initial code changes; instead the invalid CSS was removed and some generic class names were integrated instead. I have added an explicit example for dynamic inline styling in the appropriate section.

PR #726, opened 2022-11-24
Improve the Focustrap for Flow<Boolean> based sections

chausknecht:
Many typical scenaries for using a focustrap are popup based components, where the focustrap should only start to work upon the opening of the popup section. The already existing function ``trapFocusWhenever``is now improved, so that the boolean condition flow now is also applied to the initial focus setting.

### DIFF
--- a/headless/src/jsMain/kotlin/dev/fritz2/headless/foundation/focus.kt
+++ b/headless/src/jsMain/kotlin/dev/fritz2/headless/foundation/focus.kt
@@ -162,7 +162,7 @@ fun Tag<HTMLElement>.setInitialFocus() {
  * @param focus This boolean value splits the enum values into two disjoint sets: values which should set a focus
  *              after all (`true`) and those who does not at all (`false`)
  */
-enum class InitialFocus(public val focus: Boolean) {
+enum class InitialFocus(val focus: Boolean) {
 
     /**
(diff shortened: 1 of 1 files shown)


headless/src/jsMain/kotlin/dev/fritz2/headless/foundation/focus.kt
```diff
@@ -219,25 +219,25 @@ fun Tag<HTMLElement>.trapFocus(restoreFocus: Boolean = true, setInitialFocus: In
 fun Tag<HTMLElement>.trapFocusWhenever(
     condition: Flow<Boolean>,
     restoreFocus: Boolean = true,
-    setInitialFocus: InitialFocus = InitialFocus.TryToSet
+    setInitialFocus: InitialFocus = TryToSet
 ) {
+    condition handledBy {
+        if (it) setInitialFocusOnDemand(setInitialFocus)
```
jwstegemann, line 225: handled by is launching a separate coroutine. If we add the if clause with onEach to the condition-flow in line 228 we could avoid this. Or am I missing something...

chausknecht: You are right. I improved the call chain.

jwstegemann (approved):

PR #732, opened 2022-12-14
Rework Documentation

chausknecht:
This PR improves the documentation in a way that it fits better to different needs.


jamowei (approved):

jamowei (approved):

jamowei (approved):

www/src/pages/docs/30_Render HTML.md
```diff
@@ -0,0 +1,1071 @@
...
+    }
+}
+```
+
+As the identity is stable, the following properties holds for the rendered items:
+- an element is only rendered *once*, so only adding or deleting items to the store will lead to changes in the UI,
+as those new elements will be added or inserted or the deleted elements will disappear from the DOM.
+- any change to an existing item will *not* change the UI anymore. 
+
+The latter is an important aspect to consider before the use of `renderEach` for entities:
```
jwstegemann, line 514: I am not sure, if this makes it clear for the user, when it might be a good idea to use one or the other version. Since most people stop reading after they first find an information, that matches their problem, this might get them on the wrong path in this case.

jwstegemann (changes requested):

www/src/pages/docs/40_StoreCreation.md
```diff
@@ -0,0 +1,728 @@
...
+recap the shape of our model: `Person` is an *entity*, so it has some stable identifier. To determine the person in
+the list, which some new interest should be added to this person's interests list, we simply can rely on the
+`Person.id`-property.
+
+So our action consists of two parts:
+- the id of the specific person, where the interest will be added to (meta-information)
+- the interest to add (information)
+
+For simple cases like this, a `Pair` is a sufficient choice to group both information. But of course you could also
+create some (data) class or any other kotlin feature that fits.
```
jwstegemann, line 364: Wouldn't you do this by deriving a Store for this particular Person and than adding the Interes there?

www/src/pages/docs/40_StoreCreation.md
```diff
@@ -0,0 +1,728 @@
...
+    button("btn") {
+        className(storedData.tracking.data.map {
+            if(it) "spinner" else ""
+        })
+        +"save"
+        clicks handledBy storedData.save
+    }
+}
+```
+The service provides you with a `Flow` representing the description of the currently running transaction or `null`.
```
jwstegemann, line 658: In the example above it is a boolean flow...

jwstegemann (changes requested):

www/src/pages/docs/50_StoreMapping.md
```diff
@@ -0,0 +1,430 @@
...
+    val date: Lens<LocalDate, String> = lensOf(LocalDate::toString, String::toLocalDate)
+}
+```
+
+Now you can use the `Formats.date` lens for deriving appropriate stores:
+
+```kotlin
+val personStore = storeOf<Person>(Person(LocalDate(1990, 1, 1)))
+
+val birthday: Store<String> = personStore.map(Person.birthday() + Formats.date)
```
jwstegemann, line 334: Is there anywhere mentioned, that you can combine two lenses with the plus operator in general?

jwstegemann (changes requested):

www/src/pages/docs/20_Fundamentals.md
```diff
@@ -0,0 +1,356 @@
...
+of data-binding for most form elements for example.
+
+As contrary, if only one aspect is connected to a store, we call this *one way data-binding*.
+
+Let's have a look to the teaser for *custom* handlers of a store: Clearing the store's state by one button click: 
+
+```kotlin
+button {
+    +"Clear"
+    clicks handledBy store.handle { "" }
```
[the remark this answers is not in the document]
chausknecht: Adding a dot is a bit too artificial imho - what about making all text uppercase?

www/src/pages/docs/20_Fundamentals.md
```diff
@@ -0,0 +1,356 @@
...
+```kotlin
+input {
+    changes.values() handledBy   store.update
+//  ^^^^^^^^^^^^^^^^ ^^^^^^^^    ^^^^^^^^^^^^
+//  use event with   connect to  reference to some handler,
+//  new value        handler     that changes the state of the store
+}
+```
+
+As you might remember we have to initialize every store with some state value. In order to pass this value reactively
```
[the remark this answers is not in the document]
chausknecht: Yes, this should be rather some kind of supplement to the fact, that a store *holds* some value and that the initial value has to be passed.

www/src/pages/docs/20_Fundamentals.md
```diff
@@ -0,0 +1,356 @@
...
+main concepts of fritz2 within the cycling flow of data.
+
+On the left side there is the store, that keeps track of the application data. On the other side there is the DOM, 
+which holds all the HTML tags and keeps track on the emitting of events. The store's data gets connected to some
+well defined node by calling the `render` extension method on the `data`-flow of the store. Every time the stored data
+changes, the new data will be applied to the rendering code, which then creates a new dom sub-tree accordingly.
+
+From within the DOM on the right side, the user (in almost all cases!) can interact with the UI and produces events,
+which then will be used to update the store's data through so called `handler`s. The latter has access to the current
+state and the new value and can then use both to create the new state.
```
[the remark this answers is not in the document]
chausknecht: "... and the value passed by the event-flow and can then..."?
It is totally ok within this chapter to simplify things! Imho that is much better, than to expose to many details or to find the best fitting general wording. For the reader it must be clear, that the value comes from the event - of course that is just a flow that can be mapped in any way.

www/src/pages/docs/30_Render HTML.md
chausknecht: What do you propose then?

www/src/pages/docs/30_Render HTML.md
```diff
@@ -0,0 +1,1071 @@
...
+        if (it) {
+            div(id = "disclosure") {
+                +"I am open!"
+            }
+        }
+    }
+}
+```
+
+### Minimize DOM Structure Changes within Reactive Updates aka Precise Rendering
```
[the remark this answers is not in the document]
chausknecht: Imho that is partly not true. All `renderEach` are optimizations to minimize DOM changes!

But it is right, that the variant for entities targets directly heterogeneous changing models.

I am not sure, how to integrate that section, without "disturbing" the natural ordering...

www/src/pages/docs/30_Render HTML.md
```diff
@@ -0,0 +1,1071 @@
...
+### Avoid Flicker Effects with Reactive Stylings
+
+To set an initial CSS class (or any other attribute) immediately (for example to avoid flicker effects caused by the delay
+of the first value becoming available on the flow), the respective attribute-method must be called twice.
+First with the static value that should be set immediately, then with the `Flow` that provides the dynamic values:
+
+```kotlin
+val visibility: Flow<String> = ...
+
+className("invisible")
```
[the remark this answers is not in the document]
chausknecht: Haven't we concluded, that we did not want to offer some convenience functions!?
The issue is closed anyways...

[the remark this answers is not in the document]
chausknecht: (Besides: As we do not have right now, this section is correct and should remain until we might get those)

www/src/pages/docs/40_StoreCreation.md
chausknecht: The documentation has not yet arrive to those topics; so we can choose between changing examples or to accept solutions, that can be solved otherways too (we could place some hint box that mentions another approach?)

www/src/pages/docs/50_StoreMapping.md
chausknecht: No!

PR #761, opened 2023-04-24
inspector based validation

### LINKED ISSUE #760: Nesting Validators
Hello,

prior 1.0.0 it was possible to nest validators by passing the inspector. The path were automatically generated correctly, because the all validators used inspectors based on the same RootInspector.

in 1.0.0 the validator implementation is:
```
value class Validation<D, T, M>(private inline val validate: (D, T?) -> List<M>) {
    operator fun invoke(data: D, metadata: T? = null): List<M> = this.validate(data, metadata)
}
```

but should be something like
```
value class Validation<D, T, M>(private inline val validate: (Inspector<D>, T?) -> List<M>) {
    operator fun invoke(data: D, metadata: T? = null): List<M> = this.validate(inspectorOf(data), metadata)
    operator fun invoke(inspector: Inspector<D>, metadata: T? = null): List<M> = this.validate(inspector, metadata)
}
```

So a Validator can pass the Inspector to another Validator, so that the path is always generated correctly.

metin-kale-cf:
### Motivation
Real world domain objects are in most cases forms a deep object hierarchy by composing dedicated types in a sensefull way, for example some `Person` consists of fields like `name` or `birthday` but also complex fields like some `Address`, which itself represents some domain aspect with basic fields.

As validation is most of the time tied to its corresponding domain type, the validators mirror the same hierarchy as the domain classes. Thus they must be composable in the same way, the domain types are composed.

This was prior to this PR not supported by the 1.0-RC releases!

### Solution

The API of the `Validation` type changes just a little: The `validate` lambda expression now provides no longer a (domain) type `D`, but an `Inspector<D>`! There are now two `invoke` functions that take as first parameter an `Inspector<D>` and as before just a `D`, which then constructs the inspector itself. So one can use the latter for the (external) call of a validation with the domain object itself and the former for calling a validator from inside another validator!

**Remark:** If you have used the recommended `validation`-factory functions, there will be no API breaking at all, as those already have used the `Inspector<D>`. So it is very unlikely that this change will break existing code!

### Example

The following composition now works:
```kotlin
// if you use the `headless` components, prefer to use the `ComponentValidationMessage` instead of handcrafting your own!
data class Message(override val path: String, val text: String) : ValidationMessage {
    override val isError: Boolean = true
}

@Lenses
data class Person(
    val name: String,
    val birthday: LocalDate,
    val address: Address // integrate complex sub-model, with its own business rules
) {

    data class ValidationMetaData(val today: LocalDate, val knownCities: Set<String>)

    companion object {
        val validate: Validation<Person, ValidationMetaData, Message> = validation { inspector, meta ->
            inspector.map(Person.name()).let { nameInspector ->
                if (nameInspector.data.isBlank()) add(Message(nameInspector.path, "Name must not be blank!"))
            }
            inspector.map(Person.birthday()).let { birthdayInspector ->
                if (birthdayInspector.data > meta!!.today)
                    add(Message(birthdayInspector.path, "Birthday must not be in the future!"))
            }
            // call validator of `Address`-sub-model and pass mapped inspector into it as data source and for
            // creating correct paths!
            // Voilà: Validator Composition achieved!
            addAll(Address.validate(inspector.map(Person.address()), meta!!.knownCities))
        }
    }
}

@Lenses
data class Address(
    val street: String,
    val city: String
) {
    companion object {
        // enforce business rules for the `Address` domain
        val validate: Validation<Address, Set<String>, Message> = validation { inspector, cities ->
            inspector.map(Address.street()).let { streetInspector ->
                if (streetInspector.data.isBlank()) add(Message(streetInspector.path, "Street must not be blank!"))
            }
            inspector.map(Address.city()).let { cityInspector ->
                if (!cities!!.contains(cityInspector.data)) add(Message(cityInspector.path, "City does not exist!"))
            }
        }
    }
}

// and then use those:
val fritz = Person(
    "", // must not be empty!
    LocalDate(1712, 1, 24),
    Address("Am Schloss", "Potsdam") // city not in known cities list, see below
)

val errors = Person.validate(
    fritz,
    Person.ValidationMetaData(
        LocalDate(1700, 1, 1), // set "today" into the past
        setOf("Berlin", "Hamburg", "Braunschweig") // remember: no Potsdam inside
    )
)

// three errors would appear:
// Message(".name", "Name must not be blank!")
// Message(".birthday", "Birthday must not be in the future!")
// Message(".address.city", "City does not exist!")
```

### Migration Guide

If you have used the `Validation.invoke` method directly, then prefer to switch to the dedicated `validation`-factories!
Inside your validation code just remove the `inspectorOf(data)` line, that you hopefully will find and change the name of the parameter to `inspector`.

If you have not used any inspector based validation code, you simple must change the field access in such way:
```kotlin
// inside validation code:
// old
data.someField

// new
inspector.data.someField
```

Also try to replace handcrafted `path` parameters of the `Message` objects by relying on the `inspector` object:
```kotlin
// old
add(SomeMessage(".someField", ...))

// new
add(SomeMessage(inspector.path, ...))
``` 

fixes #760

### DIFF
--- a/core/src/commonMain/kotlin/dev/fritz2/validation/validation.kt
+++ b/core/src/commonMain/kotlin/dev/fritz2/validation/validation.kt
@@ -16,6 +16,48 @@ import kotlin.jvm.JvmInline
  * classes in the `commonMain` section of your Kotlin multiplatform project.
  * So you can write the validation logic once and use them on the *JS* and *JVM* side.
  *
+ * For example:
+ * ```kotlin
+ * data class Person(val name: String, val birthday: LocalDate) {
+ *     companion object {
+ *          // define validator inside of its corresponding domain type
+ *          val validate: Validator<Person, LocalDate, SomeMessage> = validation { inspector, today ->
+ *              inspector.map(Person.name()).let { nameInspector ->
+ *                  if(nameInspector.data.isBlank())
+ *                      add(SomeMessage(nameInspector.path, "Name must not be blank"))
+ *              }
+ *              inspector.map(Person.birthday()).let { birthdayInspector ->
+ *                  if(birthdayInspector.data > today!!)
+ *                      add(SomeMessage(birthdayInspector, path, "Birthday must not be in the future"))
+ *              }
+ *          }
+ *     }
+ * }
+ * ```
+ *
+ * You can also compose validators:
+ * ```kotlin
+ * data class Person(val name: String, val birthday: LocalDate) {
+ *      // take from example above!
+ * }
+ *
+ * data class User(val nickname: String, val person: Person) {
+ *      data class UserMetaData(val nicknameRepo: NicknameRepository, val today: LocalDate)
+ *
+ *      companion object {
+ *          val validate: Validator<User, UserMetaData, SomeMessage> = validation { inspector, meta ->
+ *              inspector.map(User.nickname()).let { nicknameInspector ->
+ *                  if(meta!!.nicknameRepo.exists(nicknameInspector.data))
+ *                      add(SomeMessage(nicknameInspector.path, "Nickname is already in use"))
+ *              }
+ *              // use validator of `Person` type by just calling the validator and passing the mapped inspector
+ *              // and of course the appropriate meta-data!
+ *              addAll(Person.validate(inspector.map(User.person()), meta!!.today))
+ *          }
+ *      }
+ * }
+ * ```
+ *
  * @param D data-model to validate
  * @param T metadata which perhaps is needed in validation process
  */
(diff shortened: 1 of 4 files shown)


chausknecht:
We definitely need some release notes ready text here.

I would consider a simple, yet expressive example in the description crucial. In the origin issue the main purpose of this PR is mentioned in the last sentence, but imho that should be emphasized much more: Enabling stand-alone working validators to compose smoothly.

To show a practical usage, we need some example. I would propose the classic `person` and `address` pattern, which we use in the documentation quite a lot in some variations.

The implementation looks good to me already, but some explicit unit tests about composition and stand alone usage should be added imho.

chausknecht:
> We definitely need some release notes ready text here.

Done

> The implementation looks good to me already, but some explicit unit tests about composition and stand alone usage should be added imho.

Also Done. I have fixed an (at least minor) issue on the way.

haukesomm (approved):

PR #762, opened 2023-04-24
Generate extension functions for Lens-Chaining to enable some fluent-style-API

### LINKED ISSUE #497: Extension Properties for generated lenses for simpler and shorter usage
The current lenses generation only generates a Lens like L.Object.property for each property.

When using deeper class structures the Lenses have to be stacked with the "+" operator, which can be very long.

For example to create a Lens for `OuterObject.innerObject.element.name` we have to use something like:

    L.OuterObject.innerObject + L.InnerObject.element + L.Element.name

My suggestion would be, that the Fritz2 Lenses Annotation Processor additionally generates extension properties like:

    val <R> Lens<R, InnerObject>.element get() = this + L.InnerObject.element
    val <R> Lens<R, Element>.name get() = this + L.Element.name

so that

    L.OuterObject.innerObject.element.name
can be used

Similar to that extension properties for the lenses, Store and Inspector interfaces could be extended with:

    val Store<OuterObject>.innerObject get() = sub(L.OuterObject.innerObject)
    val Inspector<OuterObject>.innerObject get() = sub(L.OuterObject.innerObject)

to be able to use the following

    OuterObjectStore.innerObject.element // type: SubStore<*, Element>
    inspector.innerObject.element // type: Inspector<Element>
    

These extensions would simplify coding and make the resulting code much more readable.

metin-kale-cf:
In order to make the combination of automatically generated lenses by the `@Lenses`-annotation more user friendly, this PR introduces a new feature: The annotation processor now creates also extension functions to enable the fluently chaining of lenses like this:

With this PR you can use:
`Model.foo().bar()`

instead of

`Model.foo() + Foo.bar()`

For more examples and explanation have a look at the [corresponding section](https://www.fritz2.dev/docs/storemapping/#combining-lenses) of the fritz2 documentation.

fixes #497

### DIFF
--- a/www/src/pages/docs/50_StoreMapping.md
+++ b/www/src/pages/docs/50_StoreMapping.md
@@ -308,11 +308,39 @@ val streetOfPerson = address + street
 
 // apply the combined lens to an example object:
 val person = Person(Address("Lerchenweg"))
-formattedBirthday.get(person) // -> "Lerchenweg"
-formattedBirthday.set("Rosenstraße") // Person(address = Address("Rosenstraße"))
+streetOfPerson.get(person) // -> "Lerchenweg"
+streetOfPerson.set("Rosenstraße") // Person(address = Address("Rosenstraße"))
 ```
 
-This feature is i.a. very useful for formatting values like you will learn about in the next section.
+Let us recap, how this example would work with automatic generated lenses:
+```kotlin
+@Lenses
+data class Address(val street: String) { companion object }
+
+@Lenses
+data class Person(val address: Address) { companion object }
+
+val streetOfPerson = Person.address() + Address.street()
+```
+
+This works, but the syntax is quite cumbersome; especially for deeper nested models!
+
+This is why our automatic `@Lenses`-annotation-processor has a dedicated support for deeper nested models as well and
+creates extension functions for all lenses, so you can *chain* the calls in a fluent way:
+
+```kotlin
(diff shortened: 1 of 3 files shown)


www/src/pages/docs/50_StoreMapping.md
```diff
@@ -308,11 +308,39 @@ val streetOfPerson = address + street
...
 // apply the combined lens to an example object:
 val person = Person(Address("Lerchenweg"))
-formattedBirthday.get(person) // -> "Lerchenweg"
-formattedBirthday.set("Rosenstraße") // Person(address = Address("Rosenstraße"))
+streetOfPerson.get(person) // -> "Lerchenweg"
+streetOfPerson.set("Rosenstraße") // Person(address = Address("Rosenstraße"))
 ```
 
-This feature is i.a. very useful for formatting values like you will learn about in the next section.
+Let us recap, how this example would work with automatic generated lenses:
```
haukesomm, line 315: Typo: `automatic` sollte hier `automatically` sein

haukesomm (changes requested):

www/src/pages/docs/50_StoreMapping.md
chausknecht: Good catch.

haukesomm (approved):

jamowei (approved):

PR #776, opened 2023-07-12
update everything for kotlin 1.9.0

jillesvangurp:
we're still on 0.14. This upgrades everything to kotlin 1.9.0.

Would you be able to cut another release for us? Also, what's the plan with the 1.0 release candidates?


Lysander:
Looks good to me!

But I could not build it yet. I hope Jan or Jens could do this. I will "trigger" them ;-)

To the plans for the 1.0 Release:
We will have a small workshop in August. After this, we will for sure release another RC with the work done.
We do not feel a hurry, since we definitely want to gather all kind of smaller issues until we decide to mark the 1.0 final.

But I am quite confident, that there won't be any big API breaking issues left. We already use the RC-* releases internally for many projects, so we regulary get feedback. 

Last but not least we would love to improve the rest of the documentation until the final 1.0 release. We have improved already the main and probably most important parts, but for the validation, routing any maybe the http sections there is some work left.

Have you already been able to toy around with the RC-Releases?

We are really happy with the projects we have used that; especially with the combination of the headless components and tailwindcss. The ui-code reads much better now and reworking the components was quite straight forward for most parts.

jillesvangurp:
I have a demo project up and running using one of the RCs: https://github.com/formation-res/kt-fullstack-demo

The headless components are a bit of an obstacle for us; just a lot of work in our application. We're not in a hurry to action that. Some way to make the migration less painful would be helpful.

jamowei (approved):

jillesvangurp:
👍 thanks!

jillesvangurp:
Ah, testing is the one thing I can't test. Should be simple to resolve

```
    Reason: Task ':core:publishJsPublicationToSonatypeRepository' uses this output of task ':core:signJvmPublication' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':core:signJvmPublication' as an input of ':core:publishJsPublicationToSonatypeRepository'.
      2. Declare an explicit dependency on ':core:signJvmPublication' from ':core:publishJsPublicationToSonatypeRepository' using Task#dependsOn.
      3. Declare an explicit dependency on ':core:signJvmPublication' from ':core:publishJsPublicationToSonatypeRepository' using Task#mustRunAfter.
 ```
 
 That's the one thing I can't change but I tracked down a potential fix here:

https://github.com/gradle/gradle/issues/17043
in this patch:
 
 https://github.com/saveourtool/save-cloud/pull/2290/files

```kotlin
 val signingTasks: TaskCollection<Sign> = tasks.withType()
    tasks.withType<AbstractPublishToMaven>().configureEach {
        dependsOn(signingTasks)
    }
```

worth a try. I can't really test this without signing keys. I pushed a version to our own maven repo this morning but would be nice to release this properly.

I gues you will have the same issue for 1.0. I would not bother with downgrading gradle, this is not likely to get fixed in gradle.

PR #778, opened 2023-08-01
Fix link to Arrow docs

serras:
The Arrow website has been updated, and now the documentation about lenses lives somewhere else


haukesomm (approved):
Thanks for the fix and have a nice weekend! ✌🏻

PR #784, opened 2023-08-17
Update to Kotlin 1.8.22 and other libraries

serras:


jamowei (approved):
Thanks!

Lysander:
There might be some places in the docs, where we refer to the kotlin-version - I think those might be adopted too.

serras:
Note that updating the Kotlin version is required for #787 to work, so the sooner this is merged into `master`, the better (if you want that PR to work, of course).

serras:
> There might be some places in the docs, where we refer to the kotlin-version - I think those might be adopted too.

The latest commit updates the references in the docs too.

Lysander:
We need some RC8-Release with PR #781 before we upgrade the Kotlin version due to some internal projects, that need this with the current Kotlin version.

I think on Monday we should be able to build the RC and after that, we can merge this one.

jamowei:
Closed in favor of #789

PR #786, opened 2023-08-17
Set up Spotless

### LINKED ISSUE #103: Enable detekt linter to fritz2 core
Add [detekt](https://detekt.github.io/detekt/) to gradle build.

serras:
Closes #103 

This sets up [Spotless](https://github.com/diffplug/spotless) using [ktlint](https://github.com/pinterest/ktlint) as back-end. Some rules like following filename rules are disabled, because the project doesn't really follow them.

Note that the PR seems big, but most of the changes are due to running `spotlessApply` to apply the linter rules consistently.


jamowei:
If I see this right, [spotless](https://github.com/diffplug/spotless) seems to be more an code formatter!?
But [detekt](https://detekt.dev/), as I know, tries to find bad code and do some analysis. Or does spotless do the same analysis and also format the code in one way? So then it would be an replacement for detekt.

@serras From your point of view is there anything against detekt or what are the key difference between detekt and spotless?

serras:
@jamowei I actually was not aware that [Detekt integrates Ktlint](https://detekt.dev/docs/rules/formatting), so with only Detekt you can have both. I'm closing this and working on #787 instead.

PR #790, opened 2023-08-23
Query parameters for GET

serras:
- `queryParameters` function which encodes a map of strings into the right form,
- additional optional argument to `get`, since query parameters often appear in such requests.


serras:
I think this should really be exposed, at least for `GET`, because the workaround of adding it by hand is quite messy.

What I think makes this solution nice is that I piggy back on JavaScript's native URL encoding functions, so there should be fewer problems there.

Lysander (approved):

PR #791, opened 2023-08-23
Serialization module

serras:
This (small) module includes a few functions which come in handy when working with fritz2's `http` module alongside `kotlinx-serialization`.


serras:
It tooks some time, but I've added an example in the module documentation.

jamowei (approved):
Nice thanks!

PR #796, opened 2023-08-30
English Overhaul of Documentation

EvaBS:
- Parts 00 - 30 have minor changes
- Part 50 needs Reviewer's attention because I changed the order of the Lens docs
- Parts 40-100 received an english text overhaul
- Changed the document order number for the www structure according to the filenames (2 were wrong)
- Removed entries from todo file which were already done

### DIFF
--- a/www/src/pages/docs/100_Web Components.md
+++ b/www/src/pages/docs/100_Web Components.md
@@ -79,7 +79,9 @@ object WeatherCard : WebComponent<HTMLDivElement>() {
         }
 }
 ```
-Also add an observed attribute named `city` to be set by the user of this component. Next, register the component and add the observed `city` to the registration.
+Also add an observed attribute named `city` to be set by the user of this component. Next, register the component and 
+add the observed `city` to the registration.
+
 ```kotlin
 fun main() {
     registerWebComponent("weather-card", WeatherCard, "city")
--- a/www/src/pages/docs/100_Web Components.md
+++ b/www/src/pages/docs/100_Web Components.md
@@ -11,10 +11,10 @@ eleventyNavigation:
 ---
 
 With fritz2, you can easily use [WebComponents](https://webcomponents.org) in any HTML-context. 
-Some of the following code-snippets are not runnable on their own. Please find the complete example 
-[here](/examples/webcomponent).
+Some of the following code-snippets are not runnable on their own. Please find the 
+[complete example here](/examples/webcomponent).
 
-### Import as script
+### Import as Script
 
 Before you can use a custom HTML element, you have to add the component to your site's scripts.
 One way is adding a `<script>` tag which points to the component in your HTML file:
@@ -51,7 +51,7 @@ Please see the [official documentation](https://kotlinlang.org/docs/js-modules.h
 for more details on this.
 
 For obvious reasons we cannot provide typesafe attributes for custom elements, 
-but you can implement a `HtmlTag` and provide an extension function for `RenderContext`:
+but you can implement an `HtmlTag` and provide an extension function for `RenderContext`:
 
 ```kotlin
 class M3Stars(job: Job, scope: Scope) : HtmlTag<HTMLElement>("m3-stars", job = job, scope = scope) {
@@ -79,18 +79,17 @@ object WeatherCard : WebComponent<HTMLDivElement>() {
         }
 }
 ```
-We also add an observed attribute named `city` which then can be set by the user of our component.
-Next, register your component:
+Also add an observed attribute named `city` to be set by the user of this component. Next, register the component and add the observed `city` to the registration.
 ```kotlin
 fun main() {
     registerWebComponent("weather-card", WeatherCard, "city")
 }
 ```
-Here you can see that we observe the `city` which must be added to the registration.
+
 
 To react to the lifecycle of your component, you can override the according methods from the specification.
 
-Packaging (i.e. as a npm-package) and publishing is out of scope of this documentation.
+Packaging (i.e. as an npm-package) and publishing is out of scope of this documentation.
 
-Again, to see it in action, please have a look at our [webcomponents example](/examples/webcomponent).
+To see it in action, please have a look at our [webcomponents example](/examples/webcomponent).
 
--- a/www/src/pages/docs/50_StoreMapping.md
+++ b/www/src/pages/docs/50_StoreMapping.md
@@ -12,55 +12,64 @@ eleventyNavigation:
 
 ## Overview
 
-fritz2 uses stores to manage the application state by holding some data model. But quite often those data does 
-not fit naturally to the needed UI-fragment. This is often due to the normalized form, where redundancy is avoided
-as much as possible. Also, HTML can only render `Strings` in the end, but the `String` representation of some data 
-types might differ from case to case, and we do not want to store those all explicitly.
+fritz2 uses stores to manage the application state by holding a data model. But quite often the data does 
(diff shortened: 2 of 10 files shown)


www/src/pages/docs/50_StoreMapping.md
```diff
@@ -12,63 +12,72 @@ eleventyNavigation:
...
 
-Like the known `map`-function from collections, where some source type `T` gets transformed to some other type `R`
-inside an expression, we can also *map* a store in order to change its source type to some other, better fitting type.
-There is one big difference between the classical `map`-function and the store's mapping functions: A store needs
-not only a function from `T -> R` (getter) but also from `R -> T` (setter) as a store manages changes!
+Remember the `map`-function from collections where a source type `T` is transformed to another type `R`
+inside of an expression? We can also *map* a store in order to change its source type to a more suitable type.
+However, there is one big difference between the classical `map`-function and the store's mapping functions: A store needs
+both a getter-function from `T -> R` and a setter-function from `R -> T` to manage changes, while the collections functions
+only requires the getter.
```
Lysander, line 27: In collections `map` that is not a "getter"; so I would drop the latter remark.

www/src/pages/docs/100_Web Components.md
```diff
@@ -79,18 +79,17 @@ object WeatherCard : WebComponent<HTMLDivElement>() {
         }
 }
 ```
-We also add an observed attribute named `city` which then can be set by the user of our component.
-Next, register your component:
+Also add an observed attribute named `city` to be set by the user of this component. Next, register the component and add the observed `city` to the registration.
```
Lysander, line 82: Respect the maximum line length please and add some line break. (That might be needed in other places as well!)

www/src/pages/docs/50_StoreMapping.md
```diff
@@ -12,63 +12,72 @@ eleventyNavigation:
...
 
-There is a universal concept in computer science for such a functionality called *lens*. You might have a look at
-the [excellent documentation on lenses](https://arrow-kt.io/learn/immutable-data/lens/) from the 
-[arrow-project](https://arrow-kt.io).
+If you are unfamiliar with the universal concept for the functionality called *lens* in computer science, take
+a look at the [excellent documentation on lenses](https://arrow-kt.io/learn/immutable-data/lens/) from the 
+[arrow-project](https://arrow-kt.io) before reading on.
 
-fritz2 also offers the method `lensOf()` for a short-and-sweet-experience, which accepts a getter- and 
+A `Lens` is basically a way to describe the relation between an outer and inner entity in a structure. It focuses on the
```
Lysander, line 35: I prefer the old structure - this is really just the teasing teaser, not the 100% in detail explanation.
In the upcoming dedicated section it is the right place to describe the details - at this place, we should omit them to focus on the big picture.

www/src/pages/docs/50_StoreMapping.md
```diff
@@ -12,63 +12,72 @@ eleventyNavigation:
...
 val hugo: Person = nameLens.set(fritz2, "hugo") // hugo = Person(1, "hugo")
 ```
-As you can see, there is no *magic*; just plain old function calling.
+As you can see, there is no magic, just plain old function calling.
 
-Let us take a step back and explore, how this concept of lenses can be used to map one store to another.
+Let's take a step back and explore how the concept of lenses can be used to map one store to another.
 
-### Mapping a Store
+### Mapping Complex Stores Using Lenses
```
Lysander, line 61: Why complex? It emphasizes some property, which is not important here!?

www/src/pages/docs/50_StoreMapping.md
```diff
@@ -311,8 +308,10 @@ val person = Person(Address("Lerchenweg"))
 streetOfPerson.get(person) // -> "Lerchenweg"
 streetOfPerson.set("Rosenstraße") // Person(address = Address("Rosenstraße"))
 ```
+This works, but the syntax is quite cumbersome - especially for deeper nested models.
```
Lysander, line 311: This syntax is not meant! The `+` Operator is cumbersome for deeper nested models ;-)
So the old appearence is valid.

www/src/pages/docs/50_StoreMapping.md
```diff
@@ -358,15 +355,9 @@ val ageLens: Lens<Person, Int> = Person.age() // cannot be used in tag attribute
 val ageLensAsString: Lens<Person, String> = Person.age().asString() // now it is usable
 ```
 
-fritz2 also provides a special `lensOf()` function for creating a `Lens<P, String>` for special types that are not
```
Lysander, line 361: Why omit the code here? Repeitition is useful here - the documentation could be read without reading the `Overview` sections! Getting the signature here is imho a fine thing.

www/src/pages/docs/50_StoreMapping.md
```diff
@@ -378,53 +369,53 @@ object Formats {
...
-| `lensForElement(index: Int): Lens<List, T>`                               | Select one element from a list by index. Useful for value objects                               |
-| `lensForElement(key: K): Lens<Map<K, V>, V>`                              | Select one element from a map by some key.                                                      |
+| `lensForElement(element: T, idProvider: IdProvider<T, I>): Lens<List, T>` | Select one element from a list of entities, therefore a stable Id is needed.                    |
+| `lensForElement(index: Int): Lens<List, T>`                               | Select one element from a list by index. Useful for value objects.                              |
+| `lensForElement(key: K): Lens<Map<K, V>, V>`                              | Select one element from a map by key.                                                           |
 
 ## Advanced Topics
 
-### Reactive Rendering of Lists of Entities with automatically Mapped Element Store
+### Reactive Rendering of Entity-Lists With Auto-Mapped Store
```
Lysander, line 408: The "element" should be kept! This is the clue of this special feature after all.

www/src/pages/docs/50_StoreMapping.md
Lysander, line 27: In collections `map` that is not a "getter"; so I would drop the latter remark.

www/src/pages/docs/100_Web Components.md
Lysander, line 82: Respect the maximum line length please and add some line break. (That might be needed in other places as well!)

www/src/pages/docs/50_StoreMapping.md
Lysander, line 35: I prefer the old structure - this is really just the teasing teaser, not the 100% in detail explanation.
In the upcoming dedicated section it is the right place to describe the details - at this place, we should omit them to focus on the big picture.

www/src/pages/docs/50_StoreMapping.md
Lysander, line 61: Why complex? It emphasizes some property, which is not important here!?

www/src/pages/docs/50_StoreMapping.md
EvaBS: I dont think this is a detail, it's what helps you understand what a lens even is.

Lysander: The base concept of the documentation is to present the user an *overview*, what is possible with fritz2. In order to get an impression of store mapping, the current *short* teasing of lenses is sufficient. It is intentionally short and hides detailed information for later explanation.

It is simply not the right way, to move those details up here! It contradicts the "new" approach of the documentation, which clearly separates the "Overview" from the "Essentials".

EvaBS: Undone ..

Lysander (approved):

PR #804, opened 2023-09-27
Removes portal stylings

### LINKED ISSUE #799: more control over z-index with modals

I have a need for nested popups/dialiogs/etc. The current implementations of modals and other things with a z-index hard code their z-index with a constant. This cannot be styled. Any styling you add for this with e.g. tailwind is simply ignored.

Add a z-index parameter to anything that uses these constants that takes an `zIndex:Int?=null` and falls back to the constant so users can override the zIndex on their modals, popups, etc.

Alternatively, let the user style this explicitly via the baseClass so that they can control this via e.g. tailwind. This currently doesn't work because of the hard coded defaults conflicting with this. 

I'm currently working around modals because they simply conflict with what I'm trying to do, which is to have fine grained control over what is on top of what. 

**Additional context**

-

### LINKED ISSUE #802: toast container styling adds display: contents, which breaks styling
**Describe the bug**

When I add a toast container, it ends up with a `portal-container` class and a styling rule that
adds this rule somehow.

```css
.portal-container {
  display: contents;
}
```

This breaks my tailwind styling. Including the styling suggested in the manual. Instead of appearing at the top on the right, the toasts appear at the bottom below my other content because the container is not rendered as a block.

I can change it to `display: block` in my browser and that fixes it. But I can't do this from code because of this undesired style rule overrides this.

It seems like #782 broke this. It was working fine before that. Right now, toasts are not usable for me.

**Expected behavior**

There are no magic classes being added to anything and I have full control over the layout. Alternatively, there is a nested div that gets added with the styling of my choice.

**Screenshots**

Lysander:
The default `z-index`-Stylings are removed entirely from the portal-factory. As the portal-containers can be defined by the user, there is no more need to "hide" them with `display-contents` and implement some `z-index` behavior for their children.

This is a huge step to consequently remain headless and be agnostic of any styling!

fixes #799
fixes #802

### DIFF
--- a/headless/src/jsMain/kotlin/dev/fritz2/headless/components/modal.kt
+++ b/headless/src/jsMain/kotlin/dev/fritz2/headless/components/modal.kt
@@ -16,9 +16,12 @@ import org.w3c.dom.*
  *
  * For more information refer to the [official documentation](https://www.fritz2.dev/headless/modal/)
  */
-class Modal : OpenClose(), WithJob {
+class Modal(id: String?) : OpenClose(), WithJob {
+
+    val componentId: String by lazy { id ?: Id.next() }
 
     override val job: Job = Job()
+
     var restoreFocus: Boolean = true
     var setInitialFocus: InitialFocus = InitialFocus.InsistToSet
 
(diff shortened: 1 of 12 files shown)


headless/src/jsMain/kotlin/dev/fritz2/headless/components/modal.kt
```diff
@@ -27,7 +30,7 @@ class Modal : OpenClose(), WithJob {
     fun init() {
         opened.filter { it }.handledBy {
             PortalRenderContext.run {
-                portal(zIndex = PORTALLING_MODAL_ZINDEX, tag = RenderContext::dialog) { close ->
+                portal("display: contents", id = componentId, tag = RenderContext::dialog) { close ->
```
haukesomm, line 33: It is generally not recommended to specify styling directly in an elements `class` attribute.
The above code results in an attribute like this: `class="display: contents"`.

Instead, the `style` attribute should be used:
```kotlin
attr("style", "display: contents;")
```

haukesomm (changes requested):

haukesomm (approved):

Issue #704, opened 2022-11-01
Investigate if we can reuse a Store's job for it's Substores


jwstegemann:
See substore.kt line 17...

Lysander:
Yes we could ;)

fixed by #806

PR #806, opened 2023-10-11
Proper management of Store Lifecycle (Jobs, Handlers and Flows)

metin-kale-cf:
Currently there is no real lifecycle management for Handlers and Jobs within Stores. This not only leads to memory leaks, but also unexpected behaviour due to endless running "handledBy" Invocations. To fix that we had to do some api breaking changes:

**RootStore-Job**
Previously RootStores created their own Job, which were never cancelled. With this PR every Sotre initialization needs a 'job: Job' parameter. 

`class RootStore<D>(initialData: D, **job: Job**,  override val id: String = Id.next())`
`fun <D> storeOf(initialData: D, **job: Job**, id: String = Id.next())`

Because the Jobs within RenderContext are already managed, we also added convenience functions, which directly use the RenderContext-Job if a Store is created with storeOf.
`fun <D> WithJob.storeOf(initialData: D, job: Job = this.job, id: String = Id.next())`

Each Store, which is created outside of a `WithJob`-Scope or directly using the Constructor needs a user managed Job. The user himself is responsible to take care of the Job Lifecycle and cancel the Job, when it is not needed anymore. _It is suggested to use the Job of the surrounding RenderContext, because that one is always properly managed._

**No global handledBy**
Previously handledBy could be called everywhere. In a WithJob-Context (e.g. a RenderContext), it used that Job, but otherwhise it created a endless Job, which was running forever within the Application Lifecycle, what might cause some unexpected sideeffects.

Therefore you can now only run handledBy within

1. WithJob Scope (a RenderContext implements WithJob)
2. Within a RootStore-Instance

The 'handledBy' functions within a RootStore is protected and can only be used within the RootStore itself, a RootStore-Scope e.g. using extension functions or apply/run is not sufficent. If you explicitely want to use the Store-Job outside the RootStore, you have to create an extension function with WithJob receiver an call that function within the RootStore wrapped with withJobContext.

Example:

    object MyStore : RootStore<String>("", Job()){
        init {
            withJobContext { myFunction() } 
        }
    }
    
    fun WithJob.myFunction() {
        flowOf("ABC") handledBy MyStore.update
    }

Alongside with this change a handledBy will also be interrupted if:

1. The Store-Job has been cancelled
2. The Consumer-Job (Job of the Scope, where handledBy was called) has been cancelled

Also the Store-Flow will be completed when the Store-Job has been cancelled to prevent further sideeffects of the store.

**Change in rendering**
When using reactive rendering e.g. using Flow<T>.render or Flow<T>.renderEach the render-operation was previously never interrupted if a new Flow Value was emitted. The render-operations where queued after each other. This behaviour has changed with this PR. When the Flow emits a new value, the current render-task will be interrupted and it will directly rerender the content with the latest value. This will improve performance but in some rare circumstances it might also cause a different behaviour.


henryB99 (approved):
Looks got to me.

Lysander:
This PR has the following open TODOs imho:
- rework the documentation to reflect the necessary changes
- improve the description to match the "realease-note"-structure and add dedicated migration guides and probably show more typical code-sections

haukesomm (approved):

Lysander:
fixes #704

PR #807, opened 2023-10-12
Update tooltip.md with hint to include portalRoot

jkapeller:
I think a simple hint to include portalRoot in order to use components using the portalling mechanism would be helpful.


Lysander:
Closed in favor of #817

PR #808, opened 2023-10-18
make ComponentValidationMessage a data class

### LINKED ISSUE #801: equals()/hashCode() for ComponentValidationMessage
ComponentValidationMessage does not implement equals()/hashCode(). Even if we have exact the same ComponentValidationMessage''s, equals() will say they are un-equal.

If we e.g. use ValidatingStore.allMessages().distinctUntilChanged(), the distinctUntilChanged will be useless, because the Messages will never be equal.

We should either use a 'data class' oder implement proper equals()/hashCode()-Functions.

metin-kale-cf:
The `ComponentValidationMessage` implements the `Message` interface, which is the building block for all validation related aspects in fritz2 and is also the default type for all `headless` components. It is definitely a `value` type from the domain perspective and thuis should be a `data class`. This makes it so much easier to filter duplicates for example.

That is why we change the type from an `open class` to a `data class`.

fixes #801


Lysander:
Proposals:
- think about "duplicate"-checks / filter as convenience function?
- create proper description (include motivation -> duplicates!) for release note

Lysander (approved):

PR #813, opened 2023-11-10
Update Github workflows to new versions

jamowei:


jillesvangurp:
would suggest adding a cache: gradle to the java action. That will speed up the builds a little.

Also, if you want to use kotlin for gh actions, checkout https://typesafegithub.github.io/github-workflows-kt/

jamowei:
> would suggest adding a cache: gradle to the java action. That will speed up the builds a little.
> 
> Also, if you want to use kotlin for gh actions, checkout https://typesafegithub.github.io/github-workflows-kt/

Thanks for the hints! Migrating to the [github-workflows-kt](https://typesafegithub.github.io/github-workflows-kt/) is planned for later ;)

Lysander (approved):

PR #823, opened 2023-11-29
Improves and enhances the reactive styling

Lysander:
needs #822 to be merged first!

- removes `addToClasses` functions. Migration: Just rename all occurrences to `className`.
- substitutes the internal class name management with a `MutableStateFlow` based approach, where the former manages all different portions of the overall class names, so those can get merged appropriately and mounted only once into the tag's `class` attribute.
- add a new `className`-function variant: This generic function accepts some `Flow[T]`, which is then be used by some lamba expression to generate the class names. Additionally the calles must also provide some initial `T`, which is used to create the
  initial class names that gets set immediately to the DOM, without needing to wait for the flow's first value.
- add unit tests for new `className`-function
- add and adapt documentation
- updates Playwright to version 1.40.1

Proposal for the release note:
- extract simple examples
- add link to the doc sections

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/core/mount.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/mount.kt
@@ -126,8 +126,9 @@ internal class MountContext<T : HTMLElement>(
     override val domNode: HTMLElement = target.domNode
     override val id = target.id
     override val baseClass = target.baseClass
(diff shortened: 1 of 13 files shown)


Lysander:
Documentation is still missing, so I reset this do draft.

metin-kale-cf (approved):

jamowei:
This PR is api-breaking please provide a migration guide for the release notes (`addToClasses` -> `className`)

jamowei (approved):

PR #826, opened 2023-12-07
Improves the validation message filtering and resets the default behaviour

Lysander:
### Motivation

Up to the `RC5 / RC6` version of fritz2, the `messages()`-function on `Store`s just matched the *exact* path. This behavior was changed by those release candidates: Right now all messages that matches the exact path, but also all messages machining the sub-path will be collected.

Our experience has shown, that this behavior change was a mistake, as the 80% use case is to collect all messages for an exact path. Most of the time one has dedicated form elements, that fit exactly one property of a model and thus should only show messages for itself. It is very rarely happening, that all messages of a sub-model should be shown all together.

This is why this PR changes again the default behavior, but also introduces a dedicated variant for collecting all sub-model messages with the new `messagesOfSubModel()`-function.

Also, we expose the `messages(filterPredicate: (M) -> Boolean)`-function, which both other functions uses internally. You can use this function apply custom filtering to the messages.

#### Overview

- use `messages()` if you want only exact path-matches
- use `messagesOfSubModel()` if you want all paths that match exactly or that share the same prefix as the receiving `Store`
- use `messages(filterPredicate: (M) -> Boolean)` if you want to apply your own, custom filter

#### Migration Guide

If you want to stick with the behavior from RC5 up to now, just rename all your `messages()` calls into `messagesOfSubModel()` and you are done.

If you have created custom filtering to get the exact match, in order to bypass the behavior, you can simply drop those now.

Reverts introduced behavior from #767 and #761

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
@@ -233,6 +233,6 @@ fun <M : ValidationMessage> Store<*>.messages(): Flow<List<M>>? = messages { mes
  * Be aware that the  filtering is based upon the correct usage of [Store.path]'s field. This can be reliably achieved
  * by using [dev.fritz2.core.Inspector]s and their mappings for creating the correct path values.
  */
-fun <M : ValidationMessage> Store<*>.messagesOfSubTree(): Flow<List<M>>? = messages { message ->
+fun <M : ValidationMessage> Store<*>.messagesOfSubModel(): Flow<List<M>>? = messages { message ->
     message.path == path || message.path.startsWith("$path.")
 }
--- a/core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
@@ -147,7 +147,7 @@ fun <D, T, M> WithJob.storeOf(
     job: Job = this.job,
     id: String = Id.next(),
 ): ValidatingStore<D, T, M> =
-    ValidatingStore(initialData, validation, metadataDefault,  job, validateAfterUpdate = true, id)
+    ValidatingStore(initialData, validation, metadataDefault, job, validateAfterUpdate = true, id)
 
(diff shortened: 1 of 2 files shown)


Lysander:
Just thinking about, whether `subModel` might be a more fitting name? So `messagesOfSubModel`?

core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
```diff
@@ -168,12 +168,14 @@ fun <D, M> WithJob.storeOf(
...
- * Finds all corresponding [ValidationMessage]s to this [Store].
+ * Finds all corresponding [ValidationMessage]s to this [Store] which satisfy the [filterPredicate]-expression.
  *
  * Be aware that the  filtering is based upon the correct usage of [Store.path]'s field. This can be reliably achieved
  * by using [dev.fritz2.core.Inspector]s and their mappings for creating the correct path values.
+ *
+ * @param filterPredicate expression to filter messages.
  */
-fun <M : ValidationMessage> Store<*>.messages(): Flow<List<M>>? =
+fun <M : ValidationMessage> Store<*>.messages(filterPredicate: (M) -> Boolean): Flow<List<M>>? =
```
jamowei, line 178: why not add a default parameter?

Lysander: In order to "remove" the parameterless variant?
The API Doc is much more expressive this way - if we chose yout approach, one would have to read until the default-value and thus its behaviour is explained...

Lysander:
> Just thinking about, whether `subModel` might be a more fitting name? So `messagesOfSubModel`?

Ok, I have renamed this.

core/src/jsMain/kotlin/dev/fritz2/validation/validation.kt
jamowei: Ok, valid point

jamowei (approved):
Also fixed the PR description ;)

PR #829, opened 2023-12-18
scope handling for portal containers

metin-kale-cf:
In #781 we added portalling, but portals currently do not inherit any scope.

In this PR portal-containers get a scope of some defined parent tags.

We have two types of portals:

Bound to the ViewPort:

- Modal
- Toast

These Portals will inherit the Scope of the PortalRoot, which inherits the Scope of its parent (usually the main render Function). 
Both the portalRoot() function and the global render() function got a scope parameter, so the user can set scopes as for every tag.

The second type of portals is bound to a reference element:

- listbox
- menu
- popOver
- tooltip

These portals inherit the scope of their reference element, to which they get aligned.

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/core/foundation.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/foundation.kt
@@ -48,6 +48,7 @@ fun render(
  * @param targetElement [HTMLElement] to mount to, default is *document.body*
  * @param override if true all child elements are removed before rendering
  * @param content [RenderContext] for rendering the data to the DOM
+ * @param scope scope for tag
  * @throws MountTargetNotFoundException if [targetElement] not found
  */
 fun render(
--- a/core/src/jsMain/kotlin/dev/fritz2/core/foundation.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/foundation.kt
@@ -25,17 +25,19 @@ class MountTargetNotFoundException(message: String) : Exception(message)
  * @param selector [query selector](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)
  * of the element to mount to
  * @param override if true all child elements are removed before rendering
+ * @param scope scope for tag
  * @param content [RenderContext] for rendering the data to the DOM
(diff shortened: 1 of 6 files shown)


core/src/jsMain/kotlin/dev/fritz2/core/foundation.kt
```diff
@@ -51,6 +53,7 @@ fun render(
 fun render(
     targetElement: HTMLElement? = document.body,
     override: Boolean = true,
+    scope: (ScopeContext.() -> Unit) = {},
```
Lysander, line 56: Kdoc is missing (compare `selector` based impl above)

headless/src/jsMain/kotlin/dev/fritz2/headless/components/toast.kt
```diff
@@ -50,7 +50,7 @@ fun <E : HTMLElement> toastContainer(
     scope: (ScopeContext.() -> Unit) = {},
     tag: TagFactory<Tag<E>>
 ) {
-    PortalRenderContext.portal(classes, id, scope, tag) {
+    PortalRenderContext.portal(classes, id, scope + PortalRenderContext.scopeContext, tag) {
```
Lysander, line 53: `scope + other` would override values from `scope` by values from `other` with the same key.
Wouldn't you expect the scope-values provided by the caller of `toastContainer` to "win"?

If so, the sequence of both operands must be switched.

If not, I think this behaviour should at least be documented then, as it behaves differently than in the rest of the framework.

headless/src/jsMain/kotlin/dev/fritz2/headless/foundation/portalling.kt
```diff
@@ -92,7 +96,7 @@ fun <C : HTMLElement> RenderContext.portal(
         PortalContainer(
             classes = classes,
             id = portalId,
-            scope = scope,
+            scope = scope + this.scope,
```
Lysander, line 99: Same as above: The sequence would override values from the `portal`-call by values from outer, exsiting scope.

Lysander (changes requested):

Lysander (approved):

PR #857, opened 2024-03-15
Add `renderTrue` and `renderFalse` convenience functions

haukesomm:
#### Motivation

With [fritz1.0-RC8](https://github.com/jwstegemann/fritz2/releases/tag/v1.0-RC8) we added a bunch of convenience-render-functions to reduce boilerplate-code within the ui-sections, such as `renderIf({/*some boolean expression */}) { ... }`. In many cases however, we have to work with boolean Flows which lead to code like this:
```kotlin
someBooleanFlow.renderIf({ it }) { 
    /* render UI */ 
}
```
This looks not as clean as it could.

#### Solution

This PR adds two convenience functions `Flow<Boolean>.renderIfTrue(...)` and `Flow<Boolean>.renderIfFalse(...)` that render a given content if a boolean Flow's value is `true` or `false` respectively.

Examples:

```kotlin
val store = storeOf(true)

store.data.renderIfTrue {
    // content is rendered if the Store's value is true.
    // If the Store's value is false, nothing is rendered.
}
```

```kotlin
val store = storeOf(false)

store.data.renderIfFalse {
    // content is rendered if the Store's value is false.
    // If the Store's value is true, nothing is rendered.
}
```

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
@@ -59,7 +59,7 @@ interface RenderContext : WithJob, WithScope {
     }
 
     /**
-     * Renders the data of a boolean [Flow] only if it's value is `true`.
+     * Renders the data of a boolean [Flow] only if its value is `true`.
      *
      * @receiver [Flow] containing the data
      * @param into target to mount content to. If not set a child div is added to the [Tag] this method is called on
@@ -78,7 +78,7 @@ interface RenderContext : WithJob, WithScope {
     }
 
     /**
-     * Renders the data of a boolean [Flow] only if it's value is `false`.
+     * Renders the data of a boolean [Flow] only if its value is `false`.
      *
      * @receiver [Flow] containing the data
      * @param into target to mount content to. If not set a child div is added to the [Tag] this method is called on
--- a/core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
@@ -59,6 +59,44 @@ interface RenderContext : WithJob, WithScope {
     }
 
     /**
+     * Renders the data of a boolean [Flow] only if it's value is `true`.
+     *
+     * @receiver [Flow] containing the data
+     * @param into target to mount content to. If not set a child div is added to the [Tag] this method is called on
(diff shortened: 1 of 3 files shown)


core/src/jsTest/kotlin/dev/fritz2/core/rendercontext.kt
```diff
@@ -96,6 +96,84 @@ class RenderContextTests {
         assertNull(div(div2))
     }
 
+    @Test
+    fun testRenderIfTrueFunction() = runTest {
```
Lysander, line 100: Adapt name

core/src/jsTest/kotlin/dev/fritz2/core/rendercontext.kt
```diff
@@ -96,6 +96,84 @@ class RenderContextTests {
...
+        val divContentAfterUpdate = document.getElementById(id)?.textContent
+        assertEquals(
+            expected = expectedContentIfFalse,
+            actual = divContentAfterUpdate,
+            message = "Content must be absent"
+        )
+    }
+
+    @Test
+    fun testRenderIfFalseFunction() = runTest {
```
Lysander, line 139: Adapt name

core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
```diff
@@ -58,6 +58,44 @@ interface RenderContext : WithJob, WithScope {
         }
     }
 
+    /**
+     * Renders the data of a boolean [Flow] only if it's value is `true`.
```
Lysander, line 62: `its` instead of `it's`!

core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
```diff
@@ -58,6 +58,44 @@ interface RenderContext : WithJob, WithScope {
...
+        into: Tag<HTMLElement>? = null,
+        content: Tag<*>.() -> Unit
+    ) {
+        renderIf(predicate = { it }, into) { _ ->
+            content()
+        }
+    }
+
+    /**
+     * Renders the data of a boolean [Flow] only if it's value is `false`.
```
Lysander, line 81: `its` instead of `it's`!

Lysander (changes requested):

Lysander (approved):

PR #859, opened 2024-04-19
Rename `classes` utility function to `joinClasses` in order to reduce naming confusion

### LINKED ISSUE #828: Rename `classes` utility function to `joinClassNames` in order to reduce naming confusion
The `classes`-utility function just joins arbitrary `String`s with classnames to one `String`. It takes care of adding aprropriate spaces and drops `null` values.

The name was meant to be short, but as it seems quite often it competes with some parameter named also `classes`, which is just passed through from higher call sites. This reads not nicely:
```kotlin
classes(classes, "... some local added classes", ...)
```

It might be better to add some suffix to express its functionality: Joining of different pieces of classnames, though `joinClassNames` or `joinClasses`?
```kotlin
joinClassNames(classes, "... some local added classes", ...)
```

We could mark the current function with `@Deprecated("renamed to ``joinClassNames()``", ReplaceWith("joinClassNames(classes)"))` in order to enable a smoother transition.

haukesomm:
This PR adds a new `joinClasses` function to replace the current `classes` function in order to minimize naming confusion.
The `classes` function gets deprecated.

### Motivation

The current name was meant to be short, but as it seems quite often it competes with parameters having the same name, which are passed in from higher call sites. This might look confusing:

```kotlin
classes(classes, "... some local added classes", ...)
```

### Migration

Use the added `joinClasses` function over the now deprecated `classes` function. This can be done automatically via IntelliJ.

Ideally, optimize your imports after replacing the respective calls.

Closes #828

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/core/foundation.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/foundation.kt
@@ -119,15 +119,38 @@ fun addGlobalStyles(css: List<String>) {
  * Joins all given [classes] strings to one html-class-attribute [String]
  * by filtering all out which are null or blank.
  */
-@Deprecated("Use joinClassNames instead", ReplaceWith("joinClasses(classes)"))
+@Deprecated("Use joinClassNames instead.", ReplaceWith("joinClasses(*classes)"))
 fun classes(vararg classes: String?): String = joinClasses(*classes)
 
 /**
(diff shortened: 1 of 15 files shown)


Lysander (changes requested):
Please watch out for usage in documentation (`*.md`) files - I found two occurences inside the headless components (`radiogroup.md` and `tabgroup.md`)

Lysander (approved):

PR #884, opened 2024-09-05
Resolves various minor issues

### LINKED ISSUE #753: Add missing Aria Attribute for Disclosure
Currently the `aria-controls` property is missing on the `disclosureButton` - it is only rendered inside the `disclosureCloseButton`, which is not enough.

### LINKED ISSUE #867: Repair Disclosures Transition in Headless Demo
The `ease-in` and `ease-out` classes are interchanged. We should fix this in order to get a smoother animation.
We can burrow the correct values from some internal component's animation.

### LINKED ISSUE #873: Move `whenever` function out of `Tag` interface
Currently, the convenience functions `T.whenever(Boolean): Flow<T?>` and `Flow<T>.whenever(Boolean): Flow<T?>` are defined inside the `Tag` interface and are thus limited to such contexts. They do not rely on any `Tag`-specific apis, though, and only act as shorthand functions for vanilla `Flow` operations.

Thus, I would like to propose lifting them out of the interface so they can be used in other places as well.

haukesomm:
This PR resolves various minor issues:

#### Fixes `aria-controls` attribute in headless disclusre
Closes #753

#### Fixes the headless disclosure's panel animation
Closes #867

#### Moves `whenever` function out of `Tag` interface
As the extension functions remain in the same package, this change should practically always be non be api-breaking.

Closes #873

### DIFF
--- a/headless/src/jsMain/kotlin/dev/fritz2/headless/components/disclosure.kt
+++ b/headless/src/jsMain/kotlin/dev/fritz2/headless/components/disclosure.kt
@@ -5,8 +5,6 @@ import dev.fritz2.headless.foundation.Aria
 import dev.fritz2.headless.foundation.OpenClose
 import dev.fritz2.headless.foundation.TagFactory
 import dev.fritz2.headless.foundation.addComponentStructureInfo
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.map
 import org.w3c.dom.HTMLButtonElement
 import org.w3c.dom.HTMLDivElement
 import org.w3c.dom.HTMLElement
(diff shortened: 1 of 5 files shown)


headless/src/jsMain/kotlin/dev/fritz2/headless/components/disclosure.kt
```diff
@@ -118,11 +116,8 @@ class Disclosure<C : HTMLElement>(tag: Tag<C>, id: String?) : Tag<C> by tag, Ope
         initialize: DisclosurePanel<CP>.() -> Unit
     ) {
         addComponentStructureInfo("disclosurePanel", this@disclosurePanel.scope, this)
-        tag(this, classes, "$componentId-panel", scope) {
-            DisclosurePanel(this).run {
-                initialize()
-                render()
-            }
+        panel = tag(this, classes, "$componentId-panel", scope) {
+            DisclosurePanel(this).run(initialize)
```
Lysander, line 120: `render` is no longer needed?

haukesomm: That's correct. In this case the `render` function only contained code to dynamically set an attribute which has been moved elsewhere, leaving the function empty. Thats why I removed it altogether.

Lysander: Now I have seen it :-)

Lysander (approved):

PR #876, opened 2024-08-19
Introduce automatic Generation of Delegating Lenses: Improves Validation Support for Sealed Class Hierarchies

Lysander:
#### Motivation

Currently the support for validating sealed class hierachies is rather limited, as our provided mechanisms wont work as the following example shows:
```kotlin

import dev.fritz2.core.Lens
import dev.fritz2.core.Lenses
import dev.fritz2.core.lensOf
import dev.fritz2.headless.validation.ComponentValidationMessage
import dev.fritz2.headless.validation.warningMessage
import dev.fritz2.validation.Validation
import dev.fritz2.validation.validation
import dev.fritz2.validation.invoke

sealed interface Product {
    val name: String

    companion object {
        // Try to implement validation for the common property `name` at a central place.
        // (Have a look at the validation code of `WebFramework` data class, which calls this central validation)

        val validation: Validation<Product, Unit, ComponentValidationMessage> = validation { inspector ->
            val name = inspector.map(Product.name()) // won't compile -> there is no Lens generated.
            // ...
        }

        fun <T : Product> createGenericValidation(): Validation<T, Unit, ComponentValidationMessage> =
            validation { inspector ->
                val name = inspector.map(T.name()) // won't compile -> generic types dont carry companion objects!
            }

        inline fun <reified T> createGenericValidationInlined(): Validation<T, Unit, ComponentValidationMessage> where T : Product =
            validation { inspector ->
                val name = inspector.map(T.name()) // won't compile -> generic types dont carry companion objects!
            }
    }
}

@Lenses
data class WebFramework(override val name: String, val technology: Technology) : Product {
    companion object {
        val validation: Validation<WebFramework, Unit, ComponentValidationMessage> = validation { inspector ->
            val technology = inspector.map(WebFramework.technology())
            if (technology.data == Technology.PHP) {
                add(technology.warningMessage("Consider a better language for this task ;-)"))
            }

            // call the validation of the parent type
            addAll(Product.validation(inspector.data))
        }
    }
}

@Lenses
data class Pizza(override val name: String, val toppings: List<String>) : Product {
    companion object {
        // analogous procedure:
        // val validation ...
    }
}

enum class Technology {
    Kotlin,
    PHP,
    FSharp,
}
```

We cannot access the generated Lenses from within the interface scope, as we cannot refer to the companion objects of the implementing classes.

We could create the needed lenses by hand of course:
```kotlin
sealed interface Product {
    val name: String

    companion object {
        // generate `Lens` by hand:
[...]

PR #886, opened 2024-09-09
Add new `@NoLens` annotation

Lysander:
The newly introduced support for lenses generation for sealed classes or interfaces (see #876) must allow to configure whether some property should be picked for lens generation or not. This heavily depends on the implementation of those. The default case is to pick a property, but sometimes a property will not be implemented as constructor property inside the child data class. For such cases it is now possible to mark such properties with the `@NoLens` annotation inside the sealed type. Such marked properties will get ignored by the lens generator, so no delegating lens will be created.

Beware that this annotation is not evaluated inside the constructor of data classes!

Imagine the following example to see `@NoLens` in action:
```kotlin
@Lenses
sealed class Framework {
	// Ignore this property for delegating lens generation.
	// The property is considered to be constant for all objects,
	//  see data class below
	@NoLens
	val ignore: String

	abstract val foo: String

	companion object
}

data class Fritz2 (
	override val foo: String,
) : Framework {
	// not part of the "data", so not possible to change at copy!
	// Because of that, we cannot define any valid lens in the sealed base,
	//  so we must mark it to exclude it for lens creation!
	override val ignore: String = "Fritz2"
}
```


core/src/commonMain/kotlin/dev/fritz2/core/lens.kt
```diff
@@ -6,6 +6,15 @@ package dev.fritz2.core
 @Target(AnnotationTarget.CLASS)
 annotation class Lenses
 
+/**
+ * Used by the fritz2 gradle-plugin to identify properties in sealed classes or interfaces, that should get ignored
+ * by the lens generation.
+ *
+ * Typical use case are const properties, that are override inside the data class body and not the ctor.
```
haukesomm, line 13: Instead of `override`, `overridden` would be the grammatically correct choice here

haukesomm (changes requested):

haukesomm (approved):

PR #895, opened 2024-09-13
Fixes two keyboard-related combobox bugs

haukesomm:
This PR fixes two bugs in the headless combobox related to keyboard navigation/selection:

- [Fixes the combobox not properly closing/resetting when navigating via keyboard](https://github.com/jwstegemann/fritz2/commit/86505d2d5d1cdaaa191f7a7a6740bf4d11f4538a)
- [Fixes possible keyboard selection in closed, lazy comboboxes](https://github.com/jwstegemann/fritz2/commit/480f31a070a678f435b42fa9db9e9c579596ac5c)


headless/src/jsMain/kotlin/dev/fritz2/headless/components/combobox.kt
```diff
@@ -615,10 +624,9 @@ class Combobox<E : HTMLElement, T>(tag: Tag<E>, id: String?) : Tag<E> by tag, Op
             hook(openDropdown)
 
 
-            Window.keydowns.mapNotNull { event ->
-                if (internalState.current.opened && shortcutOf(event) == Keys.Tab) false
-                else null
-            } handledBy internalState.setOpened
+            Window.keydowns.filter { event ->
```
Lysander, line 627: I would suggest to use the filtering of `keydownsIf()`-function to apply the filtering *immediately* and just at the time the first value of the `Flow`  is collected.

You might check the whole component for similar keyboard event handlings? It is always a good idea to prefer the immediate setup of filtering and maybe behaviour controlling.

Lysander (changes requested):

headless/src/jsMain/kotlin/dev/fritz2/headless/components/combobox.kt
haukesomm: Thanks for the heads-up! I implemented your feedback and identified one other occurrance that I was able to adjust. The two individual commits have been merged and the PR description has been adjusted.

headless/src/jsMain/kotlin/dev/fritz2/headless/components/combobox.kt
```diff
@@ -551,21 +558,25 @@ class Combobox<E : HTMLElement, T>(tag: Tag<E>, id: String?) : Tag<E> by tag, Op
...
+
         private fun handleKeyboardSelections() {
-            val selectionKeydowns = keydowns {
-                if (shortcutOf(key) in setOf(Keys.ArrowUp, Keys.ArrowDown, Keys.Enter, Keys.Home, Keys.End))
-                    preventDefault()
+            val selectShortcuts = internalState.data.map { it.opened }.distinctUntilChanged().flatMapLatest { opened ->
+                if (opened) {
+                    keydownsIf {
+                        shortcutOf(this) in (itemActivationKeys + Keys.Enter)
+                    }.map { event ->
```
Lysander, line 568: This is not good, as the `preventDefault` call should be made inside the `keydownsIf` function in order to prevent fast enough.

Lysander (changes requested):

Lysander (approved):

PR #908, opened 2024-10-10
Allows manual computation of PopUpPanel's position, fixes initial positioning of the combobox's dropdown

haukesomm:
This PR consists of two changes:
1) Adds a method to manually (re-)compute the position of a PopUpPanel's dropdown if needed
2) Fixes a bug where the combobox's dropdown appears off-position when the dropdown is aligned at the end of its reference andhas a changing size


headless/src/jsMain/kotlin/dev/fritz2/headless/components/combobox.kt
```diff
@@ -798,6 +801,16 @@ class Combobox<E : HTMLElement, T>(tag: Tag<E>, id: String?) : Tag<E> by tag, Op
 
             attr(Aria.activedescendant, activeIndexStore.data.map { it?.let(::itemId) })
 
+            // This manual computation is needed due to a bug in floating-ui preventing us from using the built-in
+            // auto-update functionality.
+            // See https://github.com/floating-ui/floating-ui/issues/1740
+            // TODO: Remove once the floating-ui auto-update bug is resolved
```
Lysander, line 807: I would prefer `FIXME` over `TODO`

Also I propose to start with the `FIXME: This manual...` and then add all of your comment. This would group the whole comment, which will become obsolete, once `floating-ui` fixes that bug:

```kotlin
            // FIXME: This manual computation is needed due to a bug in floating-ui preventing us from using the built-in
            //  auto-update functionality.
            //  See https://github.com/floating-ui/floating-ui/issues/1740
            //  Remove once the floating-ui auto-update bug is resolved
```
(You might wanna change the line breaks - but beware the extra spaces that group the comment lines)

headless/src/jsMain/kotlin/dev/fritz2/headless/foundation/PopUpPanel.kt
```diff
@@ -302,14 +302,30 @@ abstract class PopUpPanel<C : HTMLElement>(
         }
     }
 
+    /**
+     * Triggers the calculation of the panel's position.
+     *
+     * The calculation process is _asynchronous_ - once the computation has finished, the resulting values are
+     * applied in the right places _internally_.
+     *
+     * In most cases it is not needed to call this method manually. There are cases, however, where this may be
```
Lysander, line 311: Sounds more fluid to me: "...there is no need to call..."

Lysander (changes requested):

PR #911, opened 2024-10-21
Improves the API doc for `renderIf`

Lysander:
Improves the API doc for `renderIf`

### DIFF
--- a/core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
+++ b/core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
@@ -39,7 +39,69 @@ interface RenderContext : WithJob, WithScope {
     }
 
     /**
-     * Renders the data of a [Flow] only if the [predicate] is true.
+     * Renders the data of a [Flow] only if the given [predicate] is true.
+     *
+     * Example:
+     * ```kotlin
+     * data class Delivery(
+     *     val state: String,
+     *     val isVisible: Boolean
+     * )
+     *
+     * val storedDelivery = storeOf(Delivery("in production", true))
+     *
+     * storedDelivery.data.renderIf(Delivery::isVisible) { delivery ->
+     *     // This whole block might be re-rendered if the state is changed! (read on)
+     *     p { +"The state of your delivery is ${delivery.state}" }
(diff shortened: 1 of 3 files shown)


core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
```diff
@@ -39,7 +39,69 @@ interface RenderContext : WithJob, WithScope {
...
+     * storedDelivery.data.renderIf(Delivery::visibility) { delivery ->
+     *     // This whole block might be re-rendered if the state is changed! (read on)
+     *     p { +"The state of your delivery is ${delivery.state}" }
+     * }
+     * ```
+     *
+     * Beware that the data value [V] and its changes are still handled reactively inside the [content] expression!
+     *
+     * So while the [predicate] may remain stably `true`, some other aspect of the model [V] may change in between.
+     * Then the whole [content]-block will be re-rendered accordingly of course!
```
haukesomm, line 62: > Then the whole [content]-block ...

Maybe use "As a result, " over "Then"?

> ... will be re-rendered ...

In this case I'd use "would" over "will" since you're using conditional case in the sentence before

core/src/jsMain/kotlin/dev/fritz2/core/rendercontext.kt
```diff
@@ -39,7 +39,69 @@ interface RenderContext : WithJob, WithScope {
...
+     *     clicks.map { "conveyed" } handledBy storedDelivery.proceed
+     * }
+     *
+     * storedDelivery.data.renderIf(Delivery::visibility) { delivery ->
+     *     // This whole block will re re-rendered on button click above!
+     *     p { +"The state of your delivery is ${delivery.state}" }
+     * }
+     * ```
+     *
+     * If you want the content to change only once, if the condition changes, thus alternates from `true` to `false`
```
haukesomm, line 84: This sentence sounds a bit off.

Maybe use something like

> If you want the content to only change once if the condition changes (i.e. switches from `true` to `false` or back) you need to rely on an appropriate [Flow]-function like ...

haukesomm (approved):

PR #915, opened 2024-11-08
Allows event propagation of headless component toggles, fixes headless popover demo

### LINKED ISSUE #897: Headless PopOver does not close when opening another overlay element
When a PopOver is opened and another component with a dropdown and a toggle element is opened via a button click as well, the first overlay is not closed as expected and stays open.

**Expected behavior**
The dropdown should be closed.

haukesomm:
This PR allows event propagation of headless component toggles.

In some headless component toggles, events were explicitly filtered by calling `stopPropagation()` and `preventDefault()` on their click events.

This caused issues when multiple overlay-components are open at the same time as overlays would not close automatically when another toggle has been clicked.

Removing the above calls enables the implementor to choose the right event handling strategy for their component but also fixes the described issues.

See https://github.com/jwstegemann/fritz2/commit/1358563c8ef5c90a15a6dd070677bb4243eb32c5

---

Additionally, a bug in the headless popOver demo is fixed where the dropdown would immediately be closed when the toggle's button was clicked.

See https://github.com/jwstegemann/fritz2/commit/40874749c2a393aa0d521cd28c4f4f2c737be2d2

---

Closes #897

### DIFF
--- a/headless-demo/src/jsMain/kotlin/dev/fritz2/headlessdemo/components/popOver.kt
+++ b/headless-demo/src/jsMain/kotlin/dev/fritz2/headlessdemo/components/popOver.kt
@@ -39,7 +39,12 @@ fun RenderContext.popOverDemo() {
             className(opened.map { if (it) "" else "text-opacity-90" })
             opened.map { if (it) "Close Popover" else "Open Popover" }.renderText()
             opened.render { isOpen ->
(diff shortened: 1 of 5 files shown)


Lysander (changes requested):
Looks good besides one aspect:

Have you checked `activations`-API? I am not sure which use case might remain to allow passing in some lambda expression. So we might remove this parameter?

haukesomm:
Good point, I'll have a look at it  and adjust the API if needed 👍🏻

Lysander (approved):

PR #921, opened 2024-11-18
Fix setting an initial value for the ComboBox

Lysander:
Currently the ComboBox dismissed an initial value. This is now fixed.

Remark: There is no test yet, that would test this behaviour. We need to extend our demo app to make this possible. We need another, dedicated issue for that.


Lysander:
For testing the issue change line 11 in combobox demo to this temporarely:
```kotlin
val preset = Country.countries.first { it.token == "DE" }
val selectionStore = storeOf<Country?>(preset)
```

haukesomm (approved):
Looks good to me!

PR #925, opened 2024-12-04
Improves support for upcasting lenses

Lysander:
Introduces a dedicated `Lens`-factory for upcasting lenses: `lensForUpcasting`

Its inner code will catch `ClassCastException`s, which can occur in nested `Mountpoints`, and throw a dedicated `CollectionLensGetException`. This solution is similar to the handling of false index access or access by keys for collections, we already have integrated.

This is necessary due to the fact, that fritz2 does not maintain any sort of `Flow`-hierarchy from the outermost to the innermost renderings. So the framework cannot control the sequence of collecting the involved flows, in order to prevent outdated data or - for this case - *types*.

This is a practical compromise, that will work very well for UIs, as the visual appearance is kind of *dictated* by the properties of the underlying type.


Lysander:
You can use this code to reproduce the problem and to test the solution:
```kotlin
// some simple model; remember to put this into some `commonMain` sourceset!
import dev.fritz2.core.Lenses

@Lenses
sealed interface ConsultationModel {
    val stockNumber: String

    companion object
}

@Lenses
data class AgencyConsultation(
    override val stockNumber: String,
    val branch: String
) : ConsultationModel {
    companion object
}

@Lenses
data class PrivateConsultation(override val stockNumber: String) : ConsultationModel {
    companion object
}

// PoC that will produce the error before the PR and prevent any error after the PR.
fun main() {
    render {
        val storedAgency = storeOf<ConsultationModel?>(null)
        val storedStockNumber = storedAgency.map(ConsultationModel.stockNumber())
        val storedBranch = storedAgency.map(ConsultationModel.agencyConsultation().branch())

        section("m-4 p-4 bg-gray-100 space-y-4") {
            h1 { +"Consultation" }

            div("flex gap-4 flex-row mb-4") {
                button("p-2 bg-gray-300") {
                    +"Agency"
                    clicks.map { AgencyConsultation("123", "Bildung") } handledBy storedAgency.update
                }
                button("p-2 bg-gray-300") {
                    +"Private"
                    clicks.map { PrivateConsultation("123") } handledBy storedAgency.update
                }
            }

            div("flex gap-4 flex-col") {
                storedAgency.data.renderNotNull {
                    label {
                        +"Stocknumber"
                        input {
                            value(storedStockNumber.data)
                            changes.values() handledBy storedStockNumber.update
                        }
                    }
                }

                storedAgency.data.onEach { delay(500) }.renderIs(AgencyConsultation::class) {
		//                         ^^^^^^^^^^
		//                         this delay will relieably produce the 
		//                         ClassCastException before PR
                    label {
                        +"Branch"
                        input {
                            value(storedBranch.data)
                            changes.values() handledBy storedBranch.update
                        }
                    }
                }
            }

        }

        section("m-4 p-4 bg-gray-100") {
            pre {
                storedAgency.data.asString().renderText(into = this)
            }
        }

        portalRoot()
    }
}
```

core/src/commonMain/kotlin/dev/fritz2/core/Lens.kt
```diff
@@ -178,6 +178,21 @@ fun <K, V> lensForElement(key: K): Lens<Map<K, V>, V> = object : Lens<Map<K, V>,
         else throw CollectionLensSetException("no item found with key='$key'")
 }
 
+/**
+ * create a [Lens] for upcasting a base (sealed) class or interface to a specific subtype.
+ */
+inline fun <P, reified C : P> lensForUpcasting(): Lens<P, C> = object : Lens<P, C> {
+    override val id: String = ""
+
+    override fun get(parent: P): C = try {
```
haukesomm, line 187: I just had another thought on this in order to avoid the try-catch - would this be a viable alternative?

```kotlin
override fun get(parent: P): C = 
    (parent as? C) ?: throw CollectionLensGetException()
```

haukesomm (approved):

haukesomm (approved):

PR #928, opened 2024-12-20
Add `mapNullable` function to `Store`s and `Inspector`s

### LINKED ISSUE #927: Add reverse equivalent for `mapNull`
Currently we have a `mapNull` function that maps a `Store<T?>` to a `Store<T>` by mapping `null` to a given default value (same for Inspectors).

I propose to add an analogous function that is capable of doing the reverse conversion from `Store<T>` to `Store<T?>` by mapping a dedicated `T` in the parent to `null` if present.

This is useful when working with a data model incapable of handling `null` values in combination with components that use a nullable DataBinding.

haukesomm:
This PR adds a `mapNullable` function to `Store`s and `Inspector`s that works as the reverse equivalent for `mapNull`.

The function can be used to map _non-nullable_ Stores/Inspectors to _nullable_ ones as follows:

```kotlin
val nonNullableStore: Store<String> = storeOf("")

val nullableStore: Store<String?> = 
    nonNullableStore.mapNullable(placeholder = "Unknown")
//                               ^^^^^^^^^^^^^^^^^^^^^^^
//                               When the parent has the specified placeholder value,
//                               the mapped Store will have `null` as its value.
```

Closes #927

### DIFF
--- a/core/src/commonMain/kotlin/dev/fritz2/core/Lens.kt
+++ b/core/src/commonMain/kotlin/dev/fritz2/core/Lens.kt
@@ -213,7 +213,6 @@ internal fun <T> mapToNonNullLens(default: T): Lens<T?, T> = object : Lens<T?, T
  *
  * @param placeholder value to be mapped to `null`
  */
-// TODO KDoc
 internal fun <T> mapToNullableLens(placeholder: T): Lens<T, T?> = object : Lens<T, T?> {
     override val id: String = ""
     override fun get(parent: T): T? = parent.takeUnless { parent == placeholder }
--- a/core/src/commonMain/kotlin/dev/fritz2/core/Lens.kt
+++ b/core/src/commonMain/kotlin/dev/fritz2/core/Lens.kt
@@ -188,14 +188,34 @@ inline fun <P, reified C : P> lensForUpcasting(): Lens<P, C> = object : Lens<P, 
 }
 
 /**
- * Creates a lens from a nullable parent to a non-nullable value using a given default-value.
+ * Creates a [Lens] from a nullable parent to a non-nullable value using the provided [default] value.
+ *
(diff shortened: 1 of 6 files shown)


core/src/commonMain/kotlin/dev/fritz2/core/Lens.kt
```diff
@@ -188,14 +188,33 @@ inline fun <P, reified C : P> lensForUpcasting(): Lens<P, C> = object : Lens<P,
...
  *
- * @param default value to be used instead of null
+ * The inverse Lens can be created using the [mapToNullableLens] factory.
+ *
+ * @param default value to be used instead of `null`
  */
-internal fun <T> defaultLens(id: String, default: T): Lens<T?, T> = object : Lens<T?, T> {
-    override val id: String = id
+internal fun <T> mapToNonNullLens(default: T): Lens<T?, T> = object : Lens<T?, T> {
+    override val id: String = ""
```
Lysander, line 201: Sure that this is a good idea? If the value of a Lens's `id` is an empty String, we do not modify the `path` like for formatting lenses.

How was this function used before? 

Does this maybe affect the validation process for generated lenses? Have you checked this?

Lysander (commented):
Please check the remarked question first.

Lysander:
I think the new function is also missing in the docs:
https://www.fritz2.dev/docs/storemapping/#summary-of-store-mapping-factories

core/src/commonMain/kotlin/dev/fritz2/core/Lens.kt
haukesomm: Good point! The parameter has never been used, though. I manually checked every occurrence and made sure it's an empty String. As the function is internal, we can be sure that it is not breaking as well 👍🏻

Lysander (approved):

PR #931, opened 2025-01-20
Stop propagation of combobox keyboard events

haukesomm:
This PR adds a missing `stopPropagation` call in the combobox's keyboard event handling in order to prevent those events from bubbling to other elements.


Lysander:
Just to make it clear: The intent is really to stop propagations upwards the DOM hierarchy and not *immediately*? (which would have been `stopImmediatePropagation`)

PR #932, opened 2025-01-22
Bumps version of Kotlin to 2.1

Lysander:
Upgrades other libs too of course.


haukesomm:
I'm closing this in favor of #938 so we can track progress there, @Lysander

PR #937, opened 2025-04-01
Improves messages of `CollectionLensGetException`

Lysander:
This PR adds a `message`-property to the `CollectionLensGetException`-type. This message property is then used inside the various `lens*`-factories for collections in order to provide some useful hints. For example the `index` or `id` is composed alongside with the name of the factory.

This is a huge improvement to trace the error!


haukesomm (approved):
Cool addition, looks fine to me!

PR #946, opened 2025-10-13
Remove KotlinPoet from Project

### LINKED ISSUE #936: Try to remove kotlinpoet from the dependencies
We should investigate, whether we can drop kotlinepoet from our dependencies to upgrade faster to new Kotlin versions.

Currently we cannot upgrade due to kotlinepoet due to #934.

As we only use it for our automatic lenses generation and those code snippets are not too long or complex, we could try to handcraft those with native string templates. If the readbility suffers not too much, I would like to drop the kotlinpoet usage.

As we have quite a good test coverage on this part, this should be not dangerous at all.

Lysander:
This PR removes the dependencies to [kotlinpoet](https://square.github.io/kotlinpoet/) from the project.

The `LensesAnnotationProcessor` internally now works without and relies on simple `StringBuilder`-functionality. The readability is sometimes worse, but also better in some places, as it resembles more the final created code structure.

Nevertheless, this is a huge step for being able to upgrade to new Kotlin versions in much quicker time!

@fixes #936


Lysander:
[Type aliases](http://kotlinlang.org/docs/type-aliases.html) are not working anymore.

We need to add some tests and repair the naming resolution for those!

Lysander:
After further investigations, it seems, that the version of `kotlinpoet` does NOT affect our upgrade schedule!

So this is probably the false approach and should be closed, if #947 will work!

Lysander:
Because of #947 we do not need this and it can be closed.

